PR #57, opened 2014-11-07
add @TAG_OFFSET@ support in versionformat for git

fasheng:
@TAG_OFFSET@ will mark the commit count since @PARENT_TAG@, and @SEQUENCE_NUMBER@ will mark the commit count since the first commit.

Well, previous pull request was closed incorrect by me, so another one is here.

PR #58, opened 2014-11-10
Fix newline at the end of @PARENT_TAG@

sttts:
@PARENT_TAG@ had a trailing newline with git as SCM, leading to invalid version strings.


aspiers:
This will cause conflicts in #57 which makes the same fix, but I'm merging anyway since I'm told it's a more urgent bug - we shouldn't delay fixing it any longer just because the other stuff in #57 is not ready yet.

PR #63, opened 2014-12-16
Use commit timestamp

jblunck:
This branch depends on fix_hgtest branch. Additionally to that I'm going to refactor the additional unittest after merging the fix_include_exclude branch.

For an enterprise setup it is important that the generated tarballs are binary identical if they are generated from the same source revision. This commit makes tar_scm use the commit timestamp as a reference for the mtime of the files included in the tarball. The unittest is using a fixed, well known value instead to check that the mtime is taken from the source repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ python:
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq bzr git mercurial subversion
-install: pip install pep8 lxml mock
+install: pip install pep8 lxml mock python-dateutil
 script: make check
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: obs-service-tar-scm
 Section: devel
 Priority: extra
 Maintainer: Daniel Gollub <dgollub@brocade.com>
-Build-Depends: debhelper (>= 8.0.0), bzr, git, mercurial, pep8, python (>= 2.6), python-argparse | python (>= 2.7), python-unittest2, subversion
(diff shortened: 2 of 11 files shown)


aspiers:
OK, other than the 'bazar' typo this looks excellent!  Is it ready to merge?

jblunck:
@aspiers Do you understand why this fails? Travis reports that python-dateutil is installed successfully ...

aspiers:
@jblunck I have no idea :-(

olafhering:
Any news on this? Not sure about the failure. looks like python-2.6 is used but later python 2.7 is referenced?

aspiers:
@olafhering Good catch - that's exactly the problem.  The bug was introduced by #73 but sadly [Travis CI didn't catch it at the time](https://travis-ci.org/openSUSE/obs-service-tar_scm/jobs/64589663); the Travis environment happened to have Python 2.7, but when it was accidentally used it worked due to lack of any external module dependencies.

aspiers:
Proposed solution in #82.  Any volunteers? :)

aspiers:
#82 is now fixed; please can you rebase this?  Thanks!

aspiers:
@jblunck Do you think you'll be able to rebase this?

aspiers:
Again, I have no clue why I didn't notice this was rebased.  Maybe there is some issue with how I receive github notifications, but it seems to work fine everywhere else.  Anyway sorry for the delay, and thanks a lot for your great contribution!

PR #67, opened 2015-03-12
Unify language in tar_scm.service

KAMiKAZOW:
Within the same file different spellings are used, e.g. "tar ball" vs. "tarball", sometimes sentences begin with a double-space, sometimes with a single space, and some sentences end with a period and others do not.

This commit removes the inconsistencies.

### DIFF
--- a/tar_scm.service
+++ b/tar_scm.service
@@ -1,5 +1,5 @@
 <service name="tar_scm">
-  <summary>Create a tar ball from SCM repository</summary>
+  <summary>Create a tarball from SCM repository</summary>
   <description>This service uses a SCM client to checkout or update from a given repository. Supported are svn, git, hg and bzr.</description>
   <param name="scm">
     <description>Used SCM</description>
@@ -14,7 +14,7 @@
     <required/>
   </param>
   <param name="subdir">
-    <description>Package just a sub directory.</description>
+    <description>Package just a subdirectory.</description>
   </param>
   <param name="version">
     <description>Specify version to be used in tarball. Defaults to automatically detected value formatted by versionformat parameter.</description>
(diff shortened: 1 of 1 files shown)


aspiers:
Thanks a lot for highlighting these inconsistencies and suggesting fixes!  Since we are being pedantic, there are one or two things I don't entirely agree with.  I'll comment on the hunks individually.

tar_scm.service
```diff
@@ -22,12 +22,12 @@
   <param name="versionformat">
     <description>
       Auto-generate version from checked out source using this format
-      string.  This parameter is used if the 'version' parameter is not
+      string. This parameter is used if the 'version' parameter is not
```
aspiers: IMHO it's better to standardize on double-spaces to end sentences, because a) it's visually clearer and b) it allows text editors to provide functionality for navigating forwards and backwards one sentence at a time.

tar_scm.service
```diff
@@ -75,28 +75,28 @@
     </description>
   </param>
   <param name="filename">
-    <description>name of package - used together with version to determine tarball name</description>
+    <description>Name of package - used together with version to determine tarball name.</description>
```
aspiers: This description is a [noun phrase](http://en.wikipedia.org/wiki/Noun_phrase), which is grammatically inconsistent with the descriptions which begin "Specify ...", since the latter are [verb phrases](http://en.wikipedia.org/wiki/Verb_phrase).  If we really care about being consistent, we should choose one type and stick to it.

KAMiKAZOW:
Just FYI: I simply used the Github web interface which does not support squashing and I simply missed some inconsistencies the first time.

aspiers:
Ah OK, I thought that might have been it due to the `patch-1` branch name :)

PR #70, opened 2015-03-20
Split up the steps of git cloning and submodule initialization

fuhbar:
Submodules in git SCM only worked when using the master branch due to
the fact that submodule initialization takes place prior to revision
selection. Therefore we need to split up the task of cloning and
submodule initialization in order to place the revision selection in
between those processes.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -72,9 +72,13 @@ def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
     """Fetch sources via git."""
     safe_run(['git', 'clone', url, clone_dir], cwd=cwd,
              interactive=sys.stdout.isatty())
+
+
+def fetch_upstream_git_submodules(clone_dir, kwargs):
+    """Recursively initialize git submodules."""
     if 'submodules' in kwargs and kwargs['submodules']:
         safe_run(['git', 'submodule', 'update', '--init', '--recursive'],
-                 clone_dir)
+                 cwd=clone_dir)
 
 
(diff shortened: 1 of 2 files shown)


aspiers:
Thanks a lot for this!  However Travis is failing because some of your lines are too long; please can you fix, run `make` to ensure it's OK, and then update the pull request via `git push -f` ?

fuhbar:
Hi Adam, fixed the PEP8 formatting issues and Travis jobs are succeeding now.

aspiers:
Thanks a lot; please can you squash these fixes into the original commit so that they don't pollute the history when I merge?

aspiers:
OK, this code change is awesome now, thanks a lot!  However we still need a corresponding regression test case.  Do you think you'd be able to add this?  There is already [existing infrastructure and tests for submodules](https://github.com/openSUSE/obs-service-tar_scm/blob/master/tests/gittests.py#L77), so it should be a relatively simple matter of adding a new test which clones and checks out a non-`master` branch before initializing the submodule.

fuhbar:
Sure, I'll write the regression test, but later on during the week. Will come back to this soon.

aspiers:
Excellent, many thanks again!

fuhbar:
Hi Adam,
eventually found some time to come back to this issue. One more organizational question regarding this: Would you like to have the tests also squashed into a single commit along the original patch or would you like to a second commit for the test?

regards,
Fabian

aspiers:
Fantastic!  The tests should be in the same commit as the changes, so please squash.  Thanks a lot!

PR #73, opened 2015-05-27
fix Makefile test runner for Debian

aspiers:
On Debian, python2 isn't in the `$PATH`, so falling back to `python` should at least get the tests running OK, even if the `#!/usr/bin/env python2` will still break:

  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=634967

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+tar_scm
 *.pyc
 .*.sw?
 tmp/
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -2,6 +2,21 @@ DESTDIR ?=
 PREFIX   = /usr/local
 SYSCFG   = /etc
 
+define first_in_path
+$(or \
+    $(firstword $(wildcard \
+        $(foreach p,$(1),$(addsuffix /$(p),$(subst :, ,$(PATH)))) \
+    )), \
+    $(error Need one of: $(1)) \
+)
+endef
+
+# On ArchLinux, /usr/bin/python is Python 3, and other distros
+# will switch to the same at various points.  So until we support
+# Python 3, we need to do our best to ensure we have Python 2.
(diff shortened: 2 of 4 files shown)


aspiers:
> plus, you should substitute $(PYTHON) into the tar_scm.py shebang.

Yeah, that's annoying but probably makes sense :-(  So you are suggesting to have a `tar_scm.py.in` and do a typical `autoconf`-style substitution in the Makefile which generates `tar_scm.py`?  That will require `.spec` file changes but I can't immediately think of a better way to ensure it works on both Debian and ArchLinux.  I'm open to other ideas.

roman-neuhauser:
yep, that's my suggestion; i don't see any other options.

roman-neuhauser:
hm, `python2.7` also looks like a viable alternative.  would that work in Debian?

aspiers:
I think it probably would - at least on my ancient Debian box here I see:

```
lrwxrwxrwx 1 root root 9 Oct 29  2012 /usr/bin/python -> python2.6*
```

roman-neuhauser:
if you need to support both 2.7 and 2.6, then i guess `$(call first_in_path,python2.7 python2.6)` (maybe with ` python2` tacked on last) would be best.

aspiers:
@roman-neuhauser How does this look now?

roman-neuhauser:
if the sed fails it'll leave an empty or partial `$(DESTDIR)$(mylibdir)/tar_scm` behind.  i'd suggest

    sed ... tar_scm.py > tar_scm.tmp
    install -m 0755 tar_scm.tmp $(DESTDIR)$(mylibdir)/tar_scm

aspiers:
How's that?

roman-neuhauser:
that looks fine, thanks!

PR #77, opened 2015-06-22
Build package for debian/ubuntu

hiberis:


adrianschroeter:
can you give a reason why you have disabled the tests? are they unfixable?

jspanjers:
See the comment: 
This is needed to make this package build on the Ubuntu
distributions at build.opensuse.org. These distributions
provide only Ubuntu's main repository, which does not contain
the mercurial package. The mercurial package is only needed
to run tests, thus disabeling the tests allows building
the package.

aspiers:
Disabling all the tests is not the correct workaround.  It would be much better to only disable the mercurial tests, and this should be extremely easy to do via a small patch to `test.py`.

hiberis:
Right! I will create a new patch to disable only the mercurial tests and create a new pull request.
Aspiers, thank you for the feedback.

aspiers:
@hiberis Excellent, many thanks! BTW, in future you don't need to close PRs - you can just update them:

http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/

PR #78, opened 2015-06-22
Fix build on Debian 7.0/8.0

hiberis:


aspiers:
Taking it on trust that this is correct - I have no idea :)

PR #79, opened 2015-07-13
Add sslverify parameter.

cfelder:
In order to checkout sources from self-signed hosts add a parameter
for disabling certificate checks in scm tools.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -68,10 +68,18 @@ def safe_run(cmd, cwd, interactive=False):
     return (proc.returncode, output)
 
 
+def is_sslverify_enabled(kwargs):
+    """Returns ``True`` if the ``sslverify`` option has been enabled or
+    not been set (default enabled) ``False`` otherwise."""
+    return 'sslverify' not in kwargs or kwargs['sslverify']
+
+
 def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
     """Fetch sources via git."""
-    safe_run(['git', 'clone', url, clone_dir], cwd=cwd,
-             interactive=sys.stdout.isatty())
+    command = ['git', 'clone', url, clone_dir]
+    if not is_sslverify_enabled(kwargs):
+        command += ['--config', 'http.sslverify=false']
+    safe_run(command, cwd=cwd, interactive=sys.stdout.isatty())
 
 
 def fetch_upstream_git_submodules(clone_dir, kwargs):
@@ -86,12 +94,17 @@ def fetch_upstream_svn(url, clone_dir, revision, cwd, kwargs):
     command = ['svn', 'checkout', '--non-interactive', url, clone_dir]
     if revision:
         command.insert(4, '-r%s' % revision)
+    if not is_sslverify_enabled(kwargs):
+        command.insert(3, '--trust-server-cert')
     safe_run(command, cwd, interactive=sys.stdout.isatty())
 
 
 def fetch_upstream_hg(url, clone_dir, revision, cwd, kwargs):
     """Fetch sources via hg."""
-    safe_run(['hg', 'clone', url, clone_dir], cwd,
+    command = ['hg', 'clone', url, clone_dir]
+    if not is_sslverify_enabled(kwargs):
+        command += ['--insecure']
+    safe_run(command, cwd,
              interactive=sys.stdout.isatty())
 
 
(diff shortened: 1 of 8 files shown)


tar_scm.py
```diff
@@ -70,8 +70,10 @@ def safe_run(cmd, cwd, interactive=False):
 
 def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
     """Fetch sources via git."""
-    safe_run(['git', 'clone', url, clone_dir], cwd=cwd,
-             interactive=sys.stdout.isatty())
+    command = ['git', 'clone', url, clone_dir]
+    if 'sslverify' in kwargs and not kwargs['sslverify']:
```
aspiers: This test is duplicated 4 times; please can you extract into a separate function to adhere to the DRY rule?

aspiers:
Thanks a lot for the submission.  Code duplication aside, I would prefer if you provided an option for verifying certificates via a different CA rather than just disabling the verification, for exactly [the same reason I wrote in this SO answer](http://stackoverflow.com/a/26785963/179332): disabling security checks should be a last resort, not the first solution offered, because that encourages a sloppy approach to security.  However if there is an option for verifying a self-signed certificate then I am in favour of also allowing disabling of verification altogether as a last resort option.

cfelder:
Thanks for your comments. You are absolutely right that in a productive environments you should not disable CA verification checks. Anyway for in-house deployments this could be an option. I am not sure whether or not all SCMs recognizes the CA's installed in the system. Then it would make sense to add your own CA to the trusted System CA's and there would be no need to enhance the API for all those SCMs supported here to trust a specified CA.

I removed the code duplication by adding a is_sslverify_enabled function.

We are using tar_scm in a in-house deployment. Therefore it would be nice to have this new option to disable the certificate checks. The default behavior has not been changed (sslverify defaults to True).

Best.
Christian

aspiers:
OK fair enough, and thanks a lot for the detailed response.  Is there any chance you could add a test for this new option?

tar_scm.py
```diff
@@ -885,6 +904,7 @@ def parse_args():
         args.submodules = True
     else:
         args.submodules = False
+    args.sslverify = True if args.sslverify == 'enable' else False
```
bmwiedemann: might be safer to reverse this logic, to only use False when there is an explicit "disable"

cfelder: I does not matter in this case because argparse already checks against choices and apply the default value if not specified. If something different than choices has been specified argparse will exit with return code 2.

cfelder:
*ping*: Are you willing to merge this?

aspiers:
Sorry, I only just saw that you added a test.  I'm certainly willing to merge, however I still have a few tiny nit-picks.  Firstly please can you squash the test into the same commit as the change?

tar_scm.py
aspiers: The current code works fine, however please reverse the logic because it's more readable that way.

tests/commontests.py
```diff
@@ -262,3 +262,9 @@ def _switch_revision_and_subdir(self, use_cache=True):
             ],
             use_cache
         )
+
+    def test_sslverify_disabled(self):
+        self.tar_scm_std('--sslverify', 'disable')
```
aspiers: Thanks a lot for adding these tests, however currently they only test that tar_scm doesn't barf when passed this option. It should also test that when the option is used, the backend SCM command is invoked with the appropriate option.  Please could you extend it to do that?  I think it should be quite easy since you can reuse existing techniques from `testassertions.py`, e.g. see how `assertRanInitialClone` tests that the initial `git clone` / `bzr checkout` etc. get run.  Thanks!

cfelder:
*ping*

tests/testassertions.py
```diff
@@ -135,6 +135,12 @@ def assertRanInitialClone(self, logpath, loglines):
         self._find(logpath, loglines,
                    self.initial_clone_command, self.update_cache_command)
 
+    def assertSSLVerifyFalse(self, logpath, loglines):
+        self._find(logpath, loglines,
+                   self.initial_clone_command +
+                   '.*' + self.sslverify_false_args,
+                   self.sslverify_false_args + 'true')
```
aspiers: This value isn't very meaningful but I recognise it's a symptom of `_find()` assuming that you want to check for something not being there, and in this case we don't.

aspiers:
Thanks a lot!

PR #80, opened 2015-08-10
git fetch remote refs.

cfelder:
Fetch remote references in order to check them out.
switch_revision_git already supports checking out refs if they are known in
your cloned repository but when using code review systems e.g. gerrit the
references will not be cloned from the origin. Therefore fetch the specified
reference and create it locally in fetch_upstream_git. Afterwards
switch_revision_git can switch to that reference.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -80,6 +80,15 @@ def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
     if not is_sslverify_enabled(kwargs):
         command += ['--config', 'http.sslverify=false']
     safe_run(command, cwd=cwd, interactive=sys.stdout.isatty())
+    if revision:
+        # check if the reference already exists.
+        try:
+            safe_run(['git', 'rev-parse', '--verify', '--quiet', revision],
+                     cwd=clone_dir, interactive=sys.stdout.isatty())
+        except SystemExit:
+            # fetch reference from url and create locally
+            safe_run(['git', 'fetch', url, revision + ':' + revision],
+                     cwd=clone_dir, interactive=sys.stdout.isatty())
 
 
 def fetch_upstream_git_submodules(clone_dir, kwargs):
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -83,21 +83,23 @@ def is_sslverify_enabled(kwargs):
     return 'sslverify' not in kwargs or kwargs['sslverify']
 
 
+def git_ref_exists(clone_dir, revision):
+    rc, _ = run_cmd(['git', 'rev-parse', '--verify', '--quiet', revision],
+                    cwd=clone_dir, interactive=sys.stdout.isatty())
+    return (rc == 0)
+
+
 def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
     """Fetch sources via git."""
     command = ['git', 'clone', url, clone_dir]
     if not is_sslverify_enabled(kwargs):
         command += ['--config', 'http.sslverify=false']
     safe_run(command, cwd=cwd, interactive=sys.stdout.isatty())
-    if revision:
-        # check if the reference already exists.
-        try:
-            safe_run(['git', 'rev-parse', '--verify', '--quiet', revision],
-                     cwd=clone_dir, interactive=sys.stdout.isatty())
-        except SystemExit:
-            # fetch reference from url and create locally
-            safe_run(['git', 'fetch', url, revision + ':' + revision],
-                     cwd=clone_dir, interactive=sys.stdout.isatty())
+    # if the reference does not exist.
+    if revision and not git_ref_exists(clone_dir, revision):
+        # fetch reference from url and create locally
+        safe_run(['git', 'fetch', url, revision + ':' + revision],
+                 cwd=clone_dir, interactive=sys.stdout.isatty())
 
 
 def fetch_upstream_git_submodules(clone_dir, kwargs):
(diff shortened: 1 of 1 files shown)


aspiers:
Thanks.  I *think* I sort of understand why you want to do this, but not completely.  Please can you give a concrete example of a repository and a corresponding revision which doesn't get fetched during a clone?

Also please note I cannot merge this until you fix the test suite.

cfelder:
I fixed the test-issue. Below is an example of a Gerrit Code Review Patch referenced by ``refs/changes/61/261/3`` which needs to be fetched first.
```bash

felder@mbadmin:tmp % git clone https://iffwww.iff.kfa-juelich.de/review/gr
Cloning into 'gr'...
remote: Counting objects: 7852, done
remote: Finding sources: 100% (7852/7852)
remote: Total 7852 (delta 4478), reused 7405 (delta 4478)
Receiving objects: 100% (7852/7852), 26.34 MiB | 8.66 MiB/s, done.
Resolving deltas: 100% (4478/4478), done.
Checking connectivity... done.
felder@mbadmin:tmp % cd gr
felder@mbadmin:gr (master) % git rev-parse --verify refs/changes/61/261/3
fatal: Needed a single revision
felder@mbadmin:gr (master) % git fetch https://iffwww.iff.kfa-juelich.de/review/gr refs/changes/61/261/3
remote: Counting objects: 9, done
remote: Finding sources: 100% (5/5)
remote: Total 5 (delta 4), reused 5 (delta 4)
Unpacking objects: 100% (5/5), done.
From https://iffwww.iff.kfa-juelich.de/review/gr
 * branch            refs/changes/61/261/3 -> FETCH_HEAD
felder@mbadmin:gr (master) % git update-ref refs/changes/61/261/3 FETCH_HEAD
felder@mbadmin:gr (master) % git rev-parse --verify refs/changes/61/261/3
12a37e46717f1c894c4ae209a7cce86033fdff13
```

tar_scm.py
```diff
@@ -72,6 +72,17 @@ def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
     """Fetch sources via git."""
     safe_run(['git', 'clone', url, clone_dir], cwd=cwd,
              interactive=sys.stdout.isatty())
+    if revision:
+        # check if the reference already exists.
+        try:
+            safe_run(['git', 'rev-parse', '--verify', '--quiet', revision],
```
aspiers: Please use `subprocess.call` here.  `safe_run` is only intended to be used when you don't expect the command to fail in normal circumstances.

aspiers:
Thanks that's extremely helpful - I understand perfectly now :)

tar_scm.py
```diff
@@ -72,6 +72,17 @@ def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
...
+                     cwd=clone_dir, interactive=sys.stdout.isatty())
+        except SystemExit:
+            # fetch reference from url and create locally
+            safe_run(['git', 'fetch', url, revision], cwd=clone_dir,
```
aspiers: If you change this to

    ['git', 'fetch', url, revision + ':' + revision]

then you can eliminate the `update-ref` command below.

tar_scm.py
cfelder: I can change it to ``subprocess.call`` resulting in some code duplication from ``safe_run`` but I just would like to let you know that you are using it in the same way in ``update_cache_hg``. I tried not to duplicate code and not to break the existing style.

Therefore I would like to know if this is really what you want me to do?

aspiers: Haha, good catch :-)  Actually I didn't write that code.  But well done for copying the existing style, even if I don't like it ;-)  Yeah I think it would be better not to use `safe_run` when the command can fail in normal circumstances.

Also I now notice that `switch_revision_git` currently needs to check if a revision exists.  If we're adding a check here, then maybe that's no longer necessary - what do you think?

cfelder: The revision check in `switch_revision_git` is slightly different because it checks against the `revs = [x + revision for x in ['origin/', '']]`. Since `fetch_upstream_git` does not call `switch_revision_git` it would be impossible to change the revision to 'origin/' + revision.

aspiers:
Thanks so much - really really close to merging now!  Just one more request: please can you extract a `git_ref_exists` function to avoid the currently duplicated code which calls `git rev-parse`?

tar_scm.py
aspiers: I'm not sure I follow you.  Why would we want to change the revision to `'origin/' + revision`?  My point was that the following sequence contains redundancy:

1. `git clone`
2. Check if we have the revision we want.
3. If not fetch it.
4. Check if we have the revision we want.
5. Switch to it.

But I suppose it could be argued that step 4 is an extra safeguard against step 3. not doing what we expect.  So I'm OK with keeping that, as long as the duplicate code for steps 2. and 4. is refactored into a separate function like I already suggested to do.

cfelder:
*ping*

aspiers:
Sorry, I didn't notice that you updated - annoyingly, github doesn't send notifications after a push.

This now looks fantastic except for one thing - the `git_ref_exists` refactoring is squashed into the `run_cmd`, and there isn't even any mention of it in the commit message.  If you can split that out into a separate commit then I think this will be ready to merge.  Thanks again!

cfelder:
Done.

aspiers:
Many thanks for your patience and hard work on this!

cfelder:
Thanks for merging my contributions. Are you planning a release v0.6.0?

aspiers:
Yes that makes sense, although we need to fix #82 first.

PR #83, opened 2015-10-08
Fixes to color and wording

jengelh:


marcus-h:
On 2015-10-08 11:04:10 -0700, jengelh wrote:
>   * Drop color codes

\e is no escape sequence in python. If you replace it
with \033 is should work, e.g.,
sys.exit('\033[0;31msome red text\033[0m')

(if you would like to keep the colors...)

jengelh:
Updated.

aspiers:
Awesome, thanks a lot!

PR #84, opened 2015-10-08
Avoid messing up tags with dashes

jengelh:
When @PARENT_TAG@ is something like "foo-7.2.1-rc2", then using
versionformat="@PARENT_TAG@.g@TAG_OFFSET@" would cause
"foo7.2.1rc2.g123" to be emitted, which is — from my side — totally undesired.


aspiers:
I can sympathise with the intention behind the change, but this seems to cause valid test failures, e.g.

    AssertionError: 'repo-2015-10-08 18:13:44 +0000.tar' != 'repo-20151008T181344.tar'

so I suspect it needs a slightly more sophisticated approach than just removing this conversion.

jengelh:
Here is an update. Not sure why travis fails, I can't make sense of what E302 is supposed to be.

marcus-h:
On 2015-10-09 07:28:38 -0700, jengelh wrote:
> Here is an update. Not sure why travis fails, I can't make sense of what E302 is supposed to be.
> 
module level functions should be separated by 2 blank lines
(just insert a newline after line 431 and 439).

aspiers:
It seems there are still `AssertionError`s in the latest Travis build, e.g.

    AssertionError: 'repo-2015-12-08 23:46:25 +0000.tar' != 'repo-20151208T234625.tar'

M0ses:
@jengelh Is this still relevant or could we just close the issue?

jengelh:
Issues don't magically disappear, so yeah, still relevant.

hramrach:
Should be resolved with #390

Issue #82, opened 2015-09-21
Travis CI runs for Python 2.6 actually test against 2.7


aspiers:
As noted in #63, the Python 2.6 Travis build is actually using `/opt/python/2.7.9/bin/python2.7` by mistake, due to #73.  Proposed solution: tweak `GNUmakefile` so that it uses the default `python` in `$PATH` *unless* that points at a Python 3.x interpreter - if it does, it should fall back on the `$(call first_in_path,$(PYTHONS))` approach.

cfelder:
Maybe someone can add something like this to the `GNUmakefile`:

```PYTHON = [ `/usr/bin/python -c "import sysconfig; print sysconfig.get_python_version()"` == "2.7" ] && echo "/usr/bin/python" || $(call first_in_path,$(PYTHONS))```

PR #86, opened 2015-10-31
GNUmakefile: Use default python in $PATH unless it is a python 3.

cfelder:
Fallback to first_in_path,$(PYTHONS) if default is python 3.

fixes: #82


aspiers:
Fantastic, many thanks!

PR #89, opened 2015-11-12
fix generating the tarball

frispete:
Since we're changing into workdir, which is essentially identical to outdir, the former code cannot work correctly.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -364,7 +364,7 @@ def create_tar(repodir, outdir, dstname, extension='tar',
     cwd = os.getcwd()
     os.chdir(workdir)
 
-    tar = tarfile.open(dstname + '.' + extension, "w")
+    tar = tarfile.open(os.path.join(outdir, dstname + '.' + extension), "w")
     try:
         tar.add(topdir, recursive=False, filter=reset)
     except TypeError:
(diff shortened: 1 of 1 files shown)


aspiers:
I don't follow your logic here.  Why do you say `workdir` is essentially identical to `outdir`?  Please can you provide a test case where the original code fails?  The assertion "the former code cannot work correctly" seems dubious to me, because `tar_scm` has been working very well for a long time already.

frispete:
Here's the _service, that is failing consistently:
https://build.opensuse.org/package/view_file/home:frispete:mpd/cantata-git/_service?expand=1
without this change

aspiers:
I can't reproduce that - `osc service disabledrun` works fine for me in a checkout of this package.  Please could you provide the exact steps you took to reproduce the problem, along with the real error?  Thanks.

frispete:
I'm sorry @aspiers, that I bother you with this. Here's what I did: in order to analyse the behaviour of this _service, I called tar_scm this way:
/usr/lib/obs/service/tar_scm -v --scm git --url git://github.com/CDrummond/cantata.git --exclude .git --filename cantata --changesgenerate enable --outdir .tmp
Problem is the relative outdir here. Will prepare a new pull request with the correct fix...

PR #90, opened 2015-11-13
outdir must be an absolute path

frispete:
in order to correctly generate the tarball later on..

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -893,6 +893,8 @@ def parse_args():
     # basic argument validation
     if not os.path.isdir(args.outdir):
         sys.exit("%s: No such directory" % args.outdir)
+    elif not os.path.isabs(args.outdir):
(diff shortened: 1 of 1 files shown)


aspiers:
Sorry everyone, somehow I missed this PR before.  I have no problem with this change, however I don't fully understand it.  @frispete Your commit message said that it's necessary "in order to correctly generate the tarball later on", but why is that exactly? Please could you update the commit message to give more detail? Thanks!

hwoarang:
@aspiers since there is no activity here, would you perhaps consider https://github.com/openSUSE/obs-service-tar_scm/pull/107 which hopefully has a better commit message?

frispete:
I vote for #107, since it is the better solution. 

With regard of realpath, I don't think, it makes a difference..

aspiers:
Merged #107 instead.

PR #93, opened 2016-02-08
Create compressed tar file, depending on extension

dg0yt:


dkgroot:
That would be a nice addition !

adrianschroeter:
I do not like to double the code. We have already generic recompress service, why is that not good enough? (Having complex services contradicts the KISS principal of unix having small tools)

dg0yt:
The current real problem is that service_recompress doesn't remove the original file at the moment, and seeing two candidate tarballs stops debtransform (https://github.com/openSUSE/obs-build/issues/240). The service_recompress source code seems to have the ```rm -f original_file``` in the end of the file, but it is no longer working on build.o.o. However, I think that unconditionally removing the original file is not very wise anyway. At least there should be some check that it comes from another service. (And the comment there refers to files from services.)

As a final remark, I don't know where to report the service_recompress issue. It does not seem to be broken in any OBS related source code archive on Github.

dg0yt:
WRT to the KISS principle: why does tar_scm have the extension parameter if doesn't create anything but tar? This parameter is what already gives the impression that tar_scm would be able to produce something different. As an OBS newcomer, you are likely to learn the hard way.
And this change here is trivial, despite the fact I needed 3 commits to get it right.

frispete:
> I do not like to double the code. We have already generic recompress service, why is that not good enough? (Having complex services contradicts the KISS principal of unix having small tools)

With all due respect, Adrian, the compression service has bitten everybody, who has fiddled with the services already, and having it in the tar_scm is reducing the complexity for the user significantly by the addition of 8 lines of obvious code. 


@dg0yt: are you prepared to adjust the docs with a follow up PR, if Adrian changes his mind?

adrianschroeter:
the broken cleanup of .tar files was a different thing which got repaired meanwhile. (this has to work anyway independend of this, so atm I do still consider this a workaround for a bug).

dg0yt:
This submission was triggered by a bug in another tool, but it is more than just a workaround: It makes a very common use case simpler. It probably even saves some IOs and CPU on the servers. IMHO it implements what some (maybe most) people expect when seeing the otherwise useless extension parameter. This relates to the fact that the command line tar tool allows for compression without piping, too, only that tar takes a special parameter. Python's tarfile has transparent support for compression *for reading*.

On the other hand, plain tars are quite uncommon IMO and probably only relevant for applying other compression tools (here by service_recompress). Plain tars having an extension other than tar are even less common.

I was expecting to be asked for adding tests for the change. I must admit that I have no python experience, so it would take some time. I won't try until the change is considered acceptable.

I would update the documentation but it is not clear to me what is considered documentation in this repository: tar_scm.service? (I already updated the o.o wiki on debtransform while dealing with the recompress issue.)

I would also recreate the change as a single commit if the number of commits is too noisy.

adrianschroeter:
Sorry, I won't take this for now, since it will also conflict with the upcoming incremental storage of git archives in OBS. We can re-visit this after OBS 2.7 release.

dg0yt:
Accepted. Thank you.

PR #95, opened 2016-03-04
obscpio implementation

adrianschroeter:
New obscpio implementation. We could maintain this also seperate as fork, but I think it makes more sense to merge.

The service package will provide three services in future instead of just tar_scm. The idea behind is to share the code as much as possible, but to describe and limit each service for it's use case.

tar_scm: should behave like before
obs_scm: similar to tar_scm, but will store cpio archive with .obscpio suffix and a .obsinfo file providing additional meta information like the git commit id.
tar: Just for creating a tar ball from a local directory (unpack .obscpio archive)

a working example (when using local build only atm) is in home:adrianSuSE build. The _service file looks like this. Note that the first service is running on the server, while the rest only running inside of the build environment.

```
osc cat home:adrianSuSE build _service

<services>
  <service name="obs_scm">
    <param name="versionformat">%ad</param>
    <param name="url">git://github.com/openSUSE/obs-build.git</param>
    <param name="scm">git</param>
  </service>
  <service name="tar" mode="buildtime">
    <param name="obsinfo">obs-build.obsinfo</param>
  </service>
  <service name="recompress" mode="buildtime">
    <param name="compression">gz</param>
    <param name="file">*.tar</param>
  </service>
  <service name="set_version" mode="buildtime"/>
</services>
```

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -356,7 +356,6 @@ def create_cpio(repodir, basename, dstname, version, commit, args):
     """
     (workdir, topdir) = os.path.split(repodir)
     extension = 'obscpio'
-    excludes = args.exclude
 
     cwd = os.getcwd()
     os.chdir(workdir)
@@ -370,7 +369,7 @@ def create_cpio(repodir, basename, dstname, version, commit, args):
 
     # transform glob patterns to regular expressions
     includes = r'|'.join([fnmatch.translate(x) for x in args.include])
-    excludes = r'|'.join([fnmatch.translate(x) for x in excludes]) or r'$.'
+    excludes = r'|'.join([fnmatch.translate(x) for x in args.exclude]) or r'$.'
 
     # add topdir without filtering for now
     for root, dirs, files in os.walk(topdir, topdown=False):
@@ -395,7 +394,7 @@ def create_cpio(repodir, basename, dstname, version, commit, args):
     ret_code = proc.wait()
     if ret_code != 0:
         sys.exit("creating the cpio archive failed!")
-    archivefile.flush()
+    archivefile.close()
 
     # write meta data
     metafile = open(os.path.join(args.outdir, basename + '.obsinfo'), "w")
@@ -404,7 +403,7 @@ def create_cpio(repodir, basename, dstname, version, commit, args):
     # metafile.write("git describe: " + + "\n")
     if commit:
         metafile.write("commit: " + commit + "\n")
-    metafile.flush()
+    metafile.close()
 
     os.chdir(cwd)
 
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -232,7 +232,7 @@ def switch_revision_git(clone_dir, revision):
             print text.rstrip()
             break
 
-    if found_revision == None:
+    if found_revision is None:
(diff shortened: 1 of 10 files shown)


adrianschroeter:
you would need also the  just comitted build script from openSUSE:Tools project for local building

aspiers:
Edited description to fix Markdown presentation.

Without commenting on the architectural aspects of this, my main reaction to this PR is that it would be better to avoid the duplication between the two `*.service` definition files.  You could use some simple templating / substitution trick in the `Makefile` so that the bits of XML which are shared between them would only exist in one place, and the `Makefile` would generate the files at build-time.

adrianschroeter:
k, the tar_scm.service and obs_scm.service are now create from the same template file. The tar.service is too far off and deserves an extra file IMHO.

test suite failures got fixed as well, but no new tests added yet.

JFYI, the package in my home project is now building in a way that it makes it possible to get installed next to the standard tar_scm for more testing.

aspiers:
Thanks, that's much better with the template.

tar_scm.py
```diff
@@ -313,14 +342,71 @@ def prep_tree_for_tar(repodir, subdir, outdir, dstname):
...
+    """
+    (workdir, topdir) = os.path.split(repodir)
+    extension = 'obscpio'
+    excludes = args.exclude
+
+    cwd = os.getcwd()
+    os.chdir(workdir)
+
+    archivefilename = os.path.join(args.outdir, dstname + '.' + extension)
+    archivefile = open(archivefilename, "w")
```
toabctl: where are you closing ```archivefile``` ?

tar_scm.py
```diff
@@ -313,14 +342,71 @@ def prep_tree_for_tar(repodir, subdir, outdir, dstname):
 METADATA_PATTERN = re.compile(r'.*/\.(bzr|git|hg|svn).*')
 
 
+def create_cpio(repodir, basename, dstname, version, commit, args):
+    """Create an OBS cpio archive of repodir in destination directory.
+    """
+    (workdir, topdir) = os.path.split(repodir)
+    extension = 'obscpio'
+    excludes = args.exclude
```
toabctl: Not needed. just use ```args.exclude``` directly in line#364

tar_scm.py
```diff
@@ -313,14 +342,71 @@ def prep_tree_for_tar(repodir, subdir, outdir, dstname):
...
+                proc.stdin.write("\n")
+
+    proc.stdin.close()
+    ret_code = proc.wait()
+    if ret_code != 0:
+        sys.exit("creating the cpio archive failed!")
+    archivefile.flush()
+
+    # write meta data
+    metafile = open(os.path.join(args.outdir, basename + '.obsinfo'), "w")
```
toabctl: where are you closing ```metafile``` ?

tar_scm.py
adrianschroeter: it is used as stdout target of popen in next line

tar_scm.py
adrianschroeter: k, done

adrianschroeter:
changes suggest by @toabctl got added and the new extra arguments in master branch got reverted again.

PR #97, opened 2016-04-03
Use correct level 2 bullet point (*)

plusky:
As described in [packaging guidelines](https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM))


aspiers:
Great catch, thanks a lot!

Issue #92, opened 2015-12-24
v-prefixed tags in git


matwey:
Hello,

I am sorry if this already has been discussed, or I missed something.
Some developers are used v-prefixed tags for versions, for instance `v0.4.5`, when `@PARENT_TAG@` is used, `tar_scm` creates an archive like the following `repo-v0.4.5.tar.gz`.
Which is incompatible with `set_version` which thinks that the version is `v0.4.5` and writes:

```
Version: v0.4.5
```

to the spec file.

hurricanehrndz:
This occurs with debian builds as well, setversion just seems to be ignored and will not run.

M0ses:
This should be fixed with the new parameters

* versionrewrite-pattern
* versionrewrite-replacement

https://github.com/openSUSE/obs-service-tar_scm/commit/960d0cebad3bbf86102991eb38c542bbf5a16df4#diff-defb394841429cd7c4b5ad43737277e1R64

PR #100, opened 2016-04-27
Strip "v" from Git Versions when using @PARENT_TAG@

Fishwaldo:
Strip the preceeding "v" from the version when using @PARENT_TAG@ to set the version on git sources. 

Relates to issue #92

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -447,7 +447,7 @@ def detect_version_git(repodir, versionformat):
 
     version = safe_run(['git', 'log', '-n1', '--date=short',
                         "--pretty=format:%s" % versionformat], repodir)[1]
-    return version_iso_cleanup(version)
+    return version_iso_cleanup(version.lstrip("v"))
 
 
 def detect_version_svn(repodir, versionformat):


aspiers:
Thanks - sorry I've only just noticed this for some reason.  Seems like a reasonably safe change which I think I'd like to merge, however I can't until:

- [ ] some corresponding tests are added
- [ ] the commit message contains an explanation of the change

Would be really grateful if you could make those changes.  Thanks again!

P.S. [this blog post](http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/) may help.

adrianschroeter:
sorry, but please not this way (we have discussed it before). There must be at absolute minimum an opt-out from this. Also hardcoding "v" is not really a good idea, it won't help the others (eg "release-" is also popoular).

It should become some option, maybe 

  <param name="stripversionprefix">v</param>

for example.

aspiers:
This is now possible via https://github.com/openSUSE/obs-service-tar_scm/pull/132/commits/960d0cebad3bbf86102991eb38c542bbf5a16df4 from #132 so I think we can close this at last.

PR #101, opened 2016-05-19
Honour "subdir" param for changesgenerate

### LINKED ISSUE #37: changesgenerate should honour subdir
If the `subdir` parameter didn't change since the last `tar_scm` run, then `changesgenerate` should only list changes affecting the subtree specified by the `subdir` parameter.

toabctl:
When the "subdir" parameter is set and changes are generated,
only use changes which affect the given "subdir".

Fixes #37

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -727,33 +727,41 @@ def write_changes(changes_filename, changes, version, author):
     shutil.move(tmp_fp.name, changes_filename)
 
 
-def detect_changes_commands_git(repodir, changes):
+def _git_log_cmd(cmd_args, repodir, subdir):
+    """ Helper function to call 'git log' with args"""
+    cmd = ['git', 'log'] + cmd_args
+    if subdir:
+        cmd += ['--', subdir]
+    return safe_run(cmd, cwd=repodir)[1]
+
+
+def detect_changes_commands_git(repodir, subdir, changes):
     """Detect changes between GIT revisions."""
     last_rev = changes['revision']
 
     if last_rev is None:
-        last_rev = safe_run(['git', 'log', '-n1', '--pretty=format:%H',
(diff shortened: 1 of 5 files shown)


tar_scm.py
```diff
@@ -727,33 +727,41 @@ def write_changes(changes_filename, changes, version, author):
...
+    if subdir:
+        dbg_msg += " (for subdir: %s)" % (subdir)
+    logging.debug(dbg_msg)
 
-    lines = safe_run(['git', 'log',
-                      '--reverse', '--no-merges', '--pretty=format:%s',
-                      "%s..%s" % (last_rev, current_rev)], repodir)[1]
+    cmd = ['git', 'log', '--reverse', '--no-merges', '--pretty=format:%s',
+           "%s..%s" % (last_rev, current_rev)]
+    if subdir:
```
aspiers: This is the third time in this function that this logic is duplicated, so I think it's justified creating a helper function which constructs the arguments to the `git log` command (whilst respecting `subdir`) and then runs the command.  That will help shrink this function which is starting to get uncomfortably long.

aspiers:
Other than the DRY nitpick, this looks great.  Thanks a lot!

tar_scm.py
toabctl: @aspiers Done

aspiers:
@toabctl Thanks but I think you slightly misread my suggestion, as there is still duplication. Every caller of `_git_cmd_subdir`:

- passes similar arguments (`['git', 'log', ...]`)
- immediately calls `safe_run`

That's why I suggested that the helper should take care of those too.

Having said that, you get extra brownie points for the unit test! ;-)

toabctl:
@aspiers better?

aspiers:
Yes, much - excellent, thanks!

PR #102, opened 2016-05-24
Allow to disable archive mode to track plain file(s)

toabctl:
Usually tar_scm creates a tarball archive containing the files from a git repo.
Now when "extension" is set to an empty string, no tarball will be created.
Instead the files (respecting the "exclude" and "include" parameters) are
directly copied. This is comparable to the "download_files" source service but
"tar_scm" can also create a changelog for the given file(s).
This is for example useful when you want to track a .spec file in git and
also want to have the changes done in git for the .spec file in your .changes
file.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -312,10 +312,64 @@ def prep_tree_for_tar(repodir, subdir, outdir, dstname):
 # skip vcs files base on this pattern
 METADATA_PATTERN = re.compile(r'.*/\.(bzr|git|hg|svn).*')
 
-
-def create_tar(repodir, outdir, dstname, extension='tar',
-               exclude=[], include=[], package_metadata=False):
+def _process_repodir_create_tar(repodir, outdir, dstname, exclude_filter_func,
+                                extension='tar', incl_patterns=[],
+                                excl_patterns=[]):
     """Create a tarball of repodir in destination directory."""
+    (workdir, topdir) = os.path.split(repodir)
+
+    def reset(tarinfo):
+        """Python 2.7 only: reset uid/gid to 0/0 (root)."""
+        tarinfo.uid = tarinfo.gid = 0
+        tarinfo.uname = tarinfo.gname = "root"
+        return tarinfo
+
+    def tar_filter(tarinfo):
+        if exclude_filter_func(tarinfo.name):
+            return None
+
+        return reset(tarinfo)
+
+    cwd = os.getcwd()
+    os.chdir(workdir)
+
+    # create the usual tarfile
+    tar = tarfile.open(os.path.join(outdir, dstname + '.' + extension), "w")
+    try:
+        tar.add(topdir, recursive=False, filter=reset)
+    except TypeError:
+        # Python 2.6 compatibility
+        tar.add(topdir, recursive=False)
+    for entry in map(lambda x: os.path.join(topdir, x), os.listdir(topdir)):
+        try:
+            tar.add(entry, filter=tar_filter)
+        except TypeError:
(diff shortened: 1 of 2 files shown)


toabctl:
@aspiers This is still WIP (no tests added yet) but I would like to get feedback if this is in any way something you would accept.

aspiers:
I'm a bit confused by the code changes in the single commit, so I'll just base my answer on the commit message.

On the one hand, I'd like to accept it, since it's clearly fulfilling a valid requirement for a useful feature.  On the other hand, I'm conflicted about accepting it, since the implicit mission statement of `tar_scm` is for building tarballs rather than individual files.

In an ideal world, the SCM-specific code which handles changelogs etc. would be shared between `tar_scm` and `download_files`, and then this feature would belong in `download_files`.  But I'm guessing you've done it this way because it's far easier to reuse the existing code inside `tar_scm` rather than reinvent those wheels in `download_files` - am I right?

One possible way of resolving the naming conflict might be to copy [the approach Adrian took with `obs_scm`](https://github.com/openSUSE/obs-service-tar_scm/pull/95) - i.e. providing a new source service but using tricks to co-host it within this repo.  Thoughts?

toabctl:
> I'm a bit confused by the code changes in the single commit, so I'll just base my answer on the commit message.
> 
> On the one hand, I'd like to accept it, since it's clearly fulfilling a valid requirement for a useful feature. On the other hand, I'm conflicted about accepting it, since the implicit mission statement of tar_scm is for building tarballs rather than individual files.

Well - the name has also ```scm``` so it's not only tarballs. Tarballs is just one part and I make it optional :-)

> In an ideal world, the SCM-specific code which handles changelogs etc. would be shared between tar_scm and download_files, and then this feature would belong in download_files. But I'm guessing you've done it this way because it's far easier to reuse the existing code inside tar_scm rather than reinvent those wheels in download_files - am I right?

yes. absolutely right.

> One possible way of resolving the naming conflict might be to copy the approach Adrian took with obs_scm - i.e. providing a new source service but using tricks to co-host it within this repo. Thoughts?

Hm. I haven't looked yet into ```obs_scm```. Maybe what I try todo is already possible with ```obs_scm``` ? I try to find out...

aspiers:
Thomas Bechtold <notifications@github.com> wrote:
> Well - the name has also ```scm``` so it's not only tarballs. Tarballs is just one part and I make it optional :-)

But then it should be called `tar_or_file_scm` or similar :-)

> > One possible way of resolving the naming conflict might be to copy the approach Adrian took with obs_scm - i.e. providing a new source service but using tricks to co-host it within this repo. Thoughts?
> 
> Hm. I haven't looked yet into ```obs_scm```. Maybe what I try todo is already possible with ```obs_scm``` ? I try to find out...

No, that's completely different.  But you should still look at the PR
since the tricks could be reused ...

toabctl:
> No, that's completely different.  But you should still look at the PR
> since the tricks could be reused ...

Ok. ```obs_scm``` looks good. But I guess I'll wait until the PR is merged before adding another PR using the same approach...

aspiers:
@toabctl It's safe to revisit this now :-)  And this repo now already had its scope extended beyond a single source service, so adding another (if that makes sense) is less controversial than before.  We might even want to rename the repo...  @adrianschroeter @M0ses What do you think about this PR?

M0ses:
@toabctl 
Thanks for your PR
 but is this still relevant or could we just close the issue?

If you missing something, I think it would be better to open an issue again (PR are welcome)

toabctl:
I don't plan to work on this. Closing.

PR #104, opened 2016-06-14
Download the latest submodule of git.

nick-wang:
When the origin git repo doesn't update the submodule,
"git submodule update --init --recursive" couldn't
get the latest commits in master, need the option
"--remote" to get the latest commits of remote
submodules.

If the upstream maintainer forget to update the submodules,
it will never get the latest commits of submodules without
"--remote".


aspiers:
Thanks!  Sounds reasonable but please fix [the test suite](https://github.com/openSUSE/obs-service-tar_scm/blob/master/TESTING.md) otherwise I can't merge it.

nick-wang:
Close the pull request by a new one with the fix of test suite, since i rebased the fix into one commit in my own repo.

PR #105, opened 2016-06-16
Add detect_changes_commands_svn to enable svn checkout to generate .changes file

zzebster:
First time submitting please let me know if there is anything wrong.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -817,6 +817,63 @@ def detect_changes_commands_git(repodir, subdir, changes):
     return changes
 
 
+def detect_changes_commands_svn(repodir, subdir, changes):
+    """Detect changes between GIT revisions."""
+    last_rev = changes['revision']
+    first_run = False
+    if subdir:
+        repodir = os.path.join(repodir, subdir)
+
+    if last_rev is None:
+        last_rev = get_svn_rev(repodir, 10)
+        logging.debug("First run get log for initial release")
+        first_run = True
+
+    current_rev = get_svn_rev(repodir, 1)
+
+    if last_rev == current_rev:
+        logging.debug("No new commits, skipping changes file generation")
+        return
+
+    if not first_run:
+        # Increase last_rev by 1 so we dont get duplication of log messages
+        last_rev = int(last_rev) + 1
+
+    logging.debug("Generating changes between %s and %s", last_rev,
+                  current_rev)
+    lines = get_svn_log(repodir, last_rev, current_rev)
+
+    changes['revision'] = current_rev
+    changes['lines'] = lines
+    return changes
+
+
+def get_svn_log(repodir, revision1, revision2):
+    new_lines = []
+
+    xml_lines = safe_run(['svn', 'log', '-r%s:%s' % (revision1,
+                         revision2), '--xml'], repodir)[1]
+    lines = re.findall(r"<msg>.*?</msg>", xml_lines, re.S)
+
+    for line in lines:
+        line = line.replace("<msg>", "").replace("</msg>", "")
+        new_lines = new_lines + line.split("\n")
+
+    return new_lines
+
+
+def get_svn_rev(repodir, num_commits):
+    revisions = safe_run(['svn', 'log', '-l%d' % num_commits, '-q',
+                         '--incremental'], cwd=repodir)[1].split('\n')
+    # remove blank entry on end
+    revisions.pop()
+    # return last entry
+    revision = revisions[-1]
+    # retrieve the revision number and remove r
+    revision = re.search(r'^r[0-9]*', revision, re.M).group().replace("r", "")
+    return revision
+
+
 def detect_changes(scm, url, repodir, outdir, subdir):
     """Detect changes between revisions."""
     changes = read_changes_revision(url, os.getcwd(), outdir)
@@ -825,6 +882,7 @@ def detect_changes(scm, url, repodir, outdir, subdir):
 
     detect_changes_commands = {
         'git': detect_changes_commands_git,
+        'svn': detect_changes_commands_svn,
     }
 
     if scm not in detect_changes_commands:
--- a/tests/gitsvntests.py
+++ b/tests/gitsvntests.py
@@ -1,0 +1,209 @@
+#!/usr/bin/env python2
+
+import os
+import textwrap
+import re
+
+from commontests import CommonTests
+from utils       import run_git, run_svn
+
+
+class GitSvnTests(CommonTests):
+
+    """Unit tests which are shared between git and svn."""
+
+    def test_changesgenerate_disabled(self):
+        self.tar_scm_std('--changesgenerate', 'disable')
+
+    def test_changesgenerate_no_servicedata(self):
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata()
+
+    def test_changesgenerate_corrupt_servicedata(self):
+        with open(os.path.join(self.pkgdir, '_servicedata'), 'w') as sd:
+            sd.write('this is not valid xml')
+        self.tar_scm_std('--changesgenerate', 'enable', should_succeed=False)
+
+    def test_changesgenerate_empty_servicedata_file(self):
+        sd = open(os.path.join(self.pkgdir, '_servicedata'), 'w')
+        sd.close()
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata()
+
+    def test_changesgenerate_empty_servicedata_element(self):
+        with open(os.path.join(self.pkgdir, '_servicedata'), 'w') as sd:
+            sd.write("<servicedata>\n</servicedata>\n")
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata()
+
+    def test_changesgenerate_no_changesrevision(self):
+        with open(os.path.join(self.pkgdir, '_servicedata'), 'w') as sd:
+            sd.write(textwrap.dedent("""\
+              <servicedata>
+                <service name="tar_scm">
+                  <param name="url">%s</param>
+                </service>
+              </servicedata>
+            """ % self.fixtures.repo_url))
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata()
+
+    def _write_changes_file(self):
+        contents = textwrap.dedent("""\
+          -------------------------------------------------------------------
+          Fri Oct  3 00:17:50 BST 2014 - %s
+
+          - 2
+
+          -------------------------------------------------------------------
+          Thu Sep 18 10:27:14 BST 2014 - %s
+
+          - 1
+        """ % (self.fixtures.user_email, self.fixtures.user_email))
+        with open(os.path.join(self.pkgdir, 'pkg.changes'), 'w') as f:
+            f.write(contents)
+        return contents
+
+    def test_changesgenerate_no_change_or_changes_file(self):
+        self._write_servicedata(2)
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata()
+
+    def test_changesgenerate_no_change_same_changes_file(self):
+        self._write_servicedata(2)
+        orig_changes = self._write_changes_file()
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata()
+
+    def test_changesgenerate_new_commit_no_changes_file(self):
+        self._write_servicedata(2)
+        self.fixtures.create_commits(1)
+        self.tar_scm_std('--changesgenerate', 'enable')
+        self._check_servicedata(revision=3)
+
+    def _new_change_entry_regexp(self, author, changes):
+        return textwrap.dedent("""\
+          ^-------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


aspiers:
Thanks a lot! This new feature certainly makes sense.  Please could you squash the change and the tests into a single commit?

tar_scm.py
```diff
@@ -761,6 +761,52 @@ def detect_changes_commands_git(repodir, subdir, changes):
...
+        last_rev = last_rev.replace("r", "")
+        logging.debug("First run get log for initial release")
+        lines = safe_run(['svn', 'log', '-r', '%s' % last_rev, '--xml'], repodir)[1]
+        lines = re.findall(r"<msg>.*?</msg>", lines, re.S)
+        for line in lines:
+            line = line.replace("<msg>", "").replace("</msg>", "")
+            logging.info("Line: %s" % line)
+            new_lines = new_lines + line.split("\n")
+
+    current_rev = safe_run(['svn', 'log', '-l1', '-q', '--incremental'], cwd=repodir)[1].split('\n')
```
aspiers: I think it would be worth extracting calculation of `current_rev` into a separate function, since this one is currently too long.

tar_scm.py
```diff
@@ -761,6 +761,52 @@ def detect_changes_commands_git(repodir, subdir, changes):
     return changes
 
 
+def detect_changes_commands_svn(repodir, changes):
+    """Detect changes between GIT revisions."""
+    last_rev = changes['revision']
+    new_lines = []
+
+    if last_rev is None:
+        last_rev = safe_run(['svn', 'log', '-l10', '-q', '--incremental'], cwd=repodir)[1].split('\n')
```
aspiers: This `if` clause could be in a separate function, since this one is currently too long.

tests/svntests.py
```diff
@@ -36,3 +36,212 @@ def test_versionformat_revision(self):
         basename = self.basename(version='foo2')
         th = self.assertTarOnly(basename)
         self.assertTarMemberContains(th, basename + '/a', '2')
+
+    def _check_servicedata(self, expected_dirents=2, revision=2):
```
aspiers: A lot of these tests are copied directly from `gittests.py`.  Please do not copy and paste large chunks of code, as that violates [the DRY rule](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  Instead, you can share tests between `gittests.py` and `svntests.py` by putting them in a separate class and then using that as a mixin from the `GitTests` and `SvnTests` classes.  However of course it's fine to keep the `svn`-specific tests in this file.

zzebster:
I have reverted the initial checkins and merged into one checkin to include the tests and split out the above recommendations into separate functions

aspiers:
Ah, sorry for not being clearer, but that's not what I meant; reverting the commits is not the right way to do it.  Hopefully this will make it clear:

* http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/

but if something doesn't make sense, please ask and I will try to help.  Thanks again!

zzebster:
OK I think I see what you were asking me to do now. Can you inform me how you would like me to proceed now to resolve the mess I seem to have created?

aspiers:
You'll need to do an interactive rebase (git rebase -i), squash all commits into the first one, and then force push to github (git push -f).

zzebster:
OK I have done the rebase and pushed the changes hope I have managed to do this correctly.

zzebster:
Ok I see it failed do I just need to push any fixes as normal or do the rebase each time.

aspiers:
You have indeed done it correctly! Thanks a lot, this is close to looking great now!

If you click through to the Travis test failure you will see that [it's failing due to PEP8 violations which your PR introduced](https://travis-ci.org/openSUSE/obs-service-tar_scm/jobs/138844205).  These are simple formatting errors which should be easy to fix.  Please can you address them by amending the existing single commit, i.e. use `git commit --amend`, or `git rebase`'s squash or fixup feature) and doing another `git push -f` ? Thanks again!

tar_scm.py
```diff
@@ -761,6 +761,72 @@ def detect_changes_commands_git(repodir, subdir, changes):
...
+    # remove blank entry on end
+    revisions.pop()
+    # return last entry
+    revision = revisions[-1]
+    # retrieve the revision number and remove r
+    revision = re.search(r'^r[0-9]*', revision, re.M).group().replace("r", "")
+    return revision
+
+
+def get_svn_current_rev(repodir):
```
aspiers: Other than comments, this function is only one byte different from `get_svn_last_rev`, so it's a [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) violation.  Please can you eliminate the duplication?  For example:

- ditch `get_svn_current_rev`
- rename `get_svn_last_rev` to `get_svn_rev`
- add an integer parameter `num_commits` to `get_svn_rev`
- change the existing calls to `get_svn_{last,current}` so that they call `get_svn_rev` with extra parameter values of `10` and `1` respectively

tests/gitsvntests.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python2
+
+import os
+import textwrap
+import re
+
+from commontests import CommonTests
+from utils       import run_git, run_svn
```
aspiers: Is this line really needed?

tests/gittests.py
```diff
@@ -7,11 +7,12 @@
 import textwrap
 
 from githgtests  import GitHgTests
+from gitsvntests  import GitSvnTests
```
aspiers: Please can you preserve the vertical indentation here for consistency?

tests/svntests.py
aspiers: Moving shared tests into a new `gitsvntests.py` file was a huge improvement, so thanks a lot for that.  But there is still quite a bit of code duplication being introduced in these helper functions.  For example this `_check_servicedata()` function is almost identical to the one in `gittests.py`.  It should be easy to move the common duplicated parts into `gitsvntests.py` and only keep the differing bits in `gittests.py` and `svntests.py`.

tests/svntests.py
```diff
@@ -36,3 +40,112 @@ def test_versionformat_revision(self):
...
+            <service name="tar_scm">
+              <param name="url">%s</param>
+              <param name="changesrevision">([0-9].*)</param>
+            </service>
+          </servicedata>""" % self.fixtures.repo_url
+        (expected, count) = re.subn('\s{2,}', '\s*', expected)
+        m = re.match(expected, sd)
+        self.assertTrue(m, "\n'%s'\n!~ /%s/" % (sd, expected))
+
+    def _write_servicedata(self, rev):
```
aspiers: Same for this function which is largely duplicated in `gittests.py`.

tests/svntests.py
```diff
@@ -36,3 +40,112 @@ def test_versionformat_revision(self):
...
+            sd.write(textwrap.dedent("""\
+              <servicedata>
+                <service name="tar_scm">
+                  <param name="url">%s</param>
+                  <param name="changesrevision">%s</param>
+                </service>
+              </servicedata>
+            """ % (self.fixtures.repo_url, rev)))
+
+    def _test_changesgenerate_new_commit_and_changes_file(self, author=None):
```
aspiers: Same for this function which is largely duplicated in `gittests.py`.

tests/svntests.py
```diff
@@ -36,3 +40,112 @@ def test_versionformat_revision(self):
...
+            textwrap.dedent("""\
+              - Update to version 0.6.5:
+                \* 3
+                \* 4
+                \* 5
+              """)
+        )
+        self._check_changes(orig_changes, expected_changes_regexp)
+
+    def test_changesgenerate_new_commit_and_changes_file_no_version(self):
```
aspiers: Same for this function which is largely duplicated in `gittests.py`.

tests/svntests.py
```diff
@@ -36,3 +40,112 @@ def test_versionformat_revision(self):
...
+            textwrap.dedent("""\
+              - Update to version 5:
+                \* 3
+                \* 4
+                \* 5
+              """)
+        )
+        self._check_changes(orig_changes, expected_changes_regexp)
+
+    def test_changesgenerate_new_commit_and_changes_file_with_subdir(self):
```
aspiers: Same for this function which is largely duplicated in `gittests.py`.

zzebster:
I am seeing this error in the build log but I didnt change this function in the tar_scm.py

TypeError: detect_version_git() takes exactly 3 non-keyword arguments (2 given)

tar_scm.py
```diff
@@ -761,6 +761,63 @@ def detect_changes_commands_git(repodir, subdir, changes):
...
+
+    for line in lines:
+        line = line.replace("<msg>", "").replace("</msg>", "")
+        new_lines = new_lines + line.split("\n")
+
+    return new_lines
+
+
+def get_svn_rev(repodir, num_commits):
+    revisions = safe_run(['svn', 'log', '-l%s' % num_commits, '-q',
```
aspiers: Doesn't it make more sense for `num_commits` to be an integer? So this would be `'-l%d'`.

tar_scm.py
```diff
@@ -761,6 +761,63 @@ def detect_changes_commands_git(repodir, subdir, changes):
...
 
+def detect_changes_commands_svn(repodir, subdir, changes):
+    """Detect changes between GIT revisions."""
+    last_rev = changes['revision']
+    first_run = False
+    if subdir:
+        repodir = os.path.join(repodir, subdir)
+
+    if last_rev is None:
+        last_rev = get_svn_rev(repodir, '10')
```
aspiers: As per below comment, `10` integer would make more sense than `'10'` string here.

tar_scm.py
```diff
@@ -761,6 +761,63 @@ def detect_changes_commands_git(repodir, subdir, changes):
...
+        logging.debug("First run get log for initial release")
+        first_run = True
+
+    current_rev = get_svn_rev(repodir, '1')
```
aspiers: Same for `1` here.

tests/gitsvntests.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def test_changesgenerate_new_commit_and_changes_file(self):
+        self._test_changesgenerate_new_commit_and_changes_file(
+            self.fixtures.user_email)
+
+    def test_changesgenerate_new_commit_and_changes_file_default_author(self):
+        self._test_changesgenerate_new_commit_and_changes_file()
+
+    def _write_servicedata(self, rev):
+        if self.scm == 'git':
```
aspiers: Thanks - having this method in the shared class is definitely an improvement, but it would be cleaner to have the bits specific to svn or git outside the shared class.  For example,

    class GitTests(GitHgTests):
        ...
        def changesrevision(rev):
            return self.sha1s('tag%d' % rev)
        ...

    class SvnTests(CommonTests):
        ...
        def changesrevision(rev):
            return rev
        ...

tests/gitsvntests.py
```diff
@@ -0,0 +1,241 @@
...
+        tar_scm_args = [
+            '--changesgenerate', 'enable',
+            '--version', '',
+            '--changesauthor', self.fixtures.user_email
+        ]
+        self.tar_scm_std(*tar_scm_args)
+
+        self._check_servicedata(revision=rev, expected_dirents=3)
+
+        if self.scm == 'git':
```
aspiers: Same here.

tests/gitsvntests.py
```diff
@@ -0,0 +1,241 @@
...
+        self._check_changes(orig_changes, expected_changes_regexp)
+
+    def test_changesgenerate_new_commit_and_changes_file_with_subdir(self):
+        self._write_servicedata(2)
+        orig_changes = self._write_changes_file()
+        self.fixtures.create_commits(3)
+        self.fixtures.create_commits(3, subdir='another_subdir')
+        rev = 8
+
+        git_tar_scm_args = [
```
aspiers: and here.

aspiers:
Thanks again!  This is *really* close to perfect now :-)  The main thing which has to be changed is the commit message.  The first line should be a brief summary of the change, the second line should be blank, and then there should be an explanation of the change.  It doesn't make sense to include any of the squash history, because those commits will never be present in public repository history anyway.

On top of that I made some more suggestions for improving the code further, but if you are getting sick of making changes then let me know ;-)

> TypeError: detect_version_git() takes exactly 3 non-keyword arguments (2 given)

Looks like you figure out that this could be solved with a rebase, due to #108.

aspiers:
BTW https://wiki.openstack.org/wiki/GitCommitMessages#Information_in_commit_messages is good reading if you haven't seen it already.

zzebster:
@aspiers I am struggling to work out how to pull the latest changes in this repo in my own branch to resolve the conflicts.

aspiers:
@zzebster It looks to me like you are not struggling and have already figured it out :-)  The new commit immediately looks better!

aspiers:
I don't have time to go through this with a fine toothcomb, but it all looks great on the surface now, and the tests give me a lot of confidence that it's right.  So I'm gonna merge, and if there are any minor breakages, we can deal with them later.  Thanks again for your great work here, and your patience and swift responses to my feedback!

zzebster:
Thanks again for all your help. Sorry It took me a while to get my head around some of this stuff.

aspiers:
No apology needed! It's a steep learning curve, and actually you were very quick :)

PR #106, opened 2016-06-17
Download the latest submodule of git.

nick-wang:
(Fix the test suite compare to the first pull request.)

When the origin git repo doesn't update the submodule,
"git submodule update --init --recursive" couldn't
get the latest commits in master, need the option
"--remote" to get the latest commits of remote
submodules.

If the upstream maintainer forget to update the submodules,
it will never get the latest commits of submodules without
"--remote".

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -113,9 +113,12 @@ def fetch_upstream_git(url, clone_dir, revision, cwd, kwargs):
 
 def fetch_upstream_git_submodules(clone_dir, kwargs):
     """Recursively initialize git submodules."""
-    if 'submodules' in kwargs and kwargs['submodules']:
+    if 'submodules' in kwargs and kwargs['submodules'] == 'enable':
         safe_run(['git', 'submodule', 'update', '--init', '--recursive'],
                  cwd=clone_dir)
+    elif 'submodules' in kwargs and kwargs['submodules'] == 'master':
+        safe_run(['git', 'submodule', 'update', '--init', '--recursive',
+                 '--remote'], cwd=clone_dir)
 
 
 def fetch_upstream_svn(url, clone_dir, revision, cwd, kwargs):
@@ -1116,11 +1119,13 @@ def parse_args():
                              DEFAULT_AUTHOR)
     parser.add_argument('--subdir', default='',
                         help='Package just a subdirectory of the sources')
-    parser.add_argument('--submodules', choices=['enable', 'disable'],
+    parser.add_argument('--submodules',
+                        choices=['enable', 'master', 'disable'],
                         default='enable',
                         help='Whether or not to include git submodules '
                              'from SCM commit log since a given parent '
-                             'revision (see changesrevision).')
+                             'revision (see changesrevision). Use '
+                             '\'master\' to fetch the latest master.')
     parser.add_argument('--sslverify', choices=['enable', 'disable'],
                         default='enable',
                         help='Whether or not to check server certificate '
@@ -1172,10 +1177,6 @@ def parse_args():
     else:
         args.package_meta = False
 
-    if args.submodules == 'enable':
-        args.submodules = True
-    else:
-        args.submodules = False
     args.sslverify = False if args.sslverify == 'disable' else True
 
     # force verbose mode in test-mode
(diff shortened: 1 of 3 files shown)


adrianschroeter:
Does this mean submodules always get the latest master independ of the commit reference of the submodule?
This would break if the sub modules are frozen to a known good state on purpose, right?
So this should be not the default, but could have an option for this...

nick-wang:
Yes, it will force to get the latest master of submodule. However, as long as the commit is accepted, it should work as expect. Otherwise, we need a patch to fix it.I think it is more common that we need to get the latest commits, or at least supply a method to do it.
(A method to do it as optional, run "git submodule update --remote" after "git submodule update --init --recursive" before tar scm. And we also need an entrance in _service of obs to enable it.)

adrianschroeter:
it is at least an incompatible change and might break current setups (think of some stable maintained branches which must not use the latest version). 
We shouldn't assume by default that upstream is careless regarding their tags. And even if they are, it is a bug which should be fixed on their side in first place. 

I don't mind to have an optional workaround if this does not work out though.

nick-wang:
Understand your concern about break the stable branches.
But sometimes maintainer of the parent module don't realize the submodules have already been changed (maybe because a bug fix). In this case, we have no way to package it via _service automatically before maintainer update the submodules first.
So i think we should give this option or at least have a way to do it.

Any suggestion on that if you agree to add an option of it? I am not an expert of _service. Do you have any devel doc of it? Or if i update the function, where can i doc it to user except only in git log. 
Thanks

adrianschroeter:
On Freitag, 17. Juni 2016, 02:42:50 CEST wrote nick-wang:
> Understand your concern about break the stable branches.
> But sometimes maintainer of the parent module don't realize the submodules have already been changed (maybe because a bug fix). In this case, we have no way to package it via _service automatically before maintainer update the submodules first.
> So i think we should give this option or at least have a way to do it.
> 
> Any suggestion on that if you agree to add an option of it?

Yes, I do agree.

> I am not an expert of _service. Do you have any devel doc of it?

The options are specified in the tar_scm.service file. Another option
is basically another parameter which can be used to call the service executable.

> Or if i update the function, where can i doc it to user except only in git log. 
> Thanks
> 
> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/openSUSE/obs-service-tar_scm/pull/106#issuecomment-226726874


-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imend�rffer, Jane Smithard, Graham Norton, HRB 21284 (AG N�rnberg) 
 
Maxfeldstra�e 5                         
90409 N�rnberg 
Germany

aspiers:
Thanks a lot for this!

I agree with @adrianschroeter that `tar_scm` should not fetch the submodule's latest `master` by default; it's very important that it should fetch the exact commit specified by the parent repository, since many maintainers of repos with submodules will only validate their repo using the exact submodule commits tracked in the parent, and switching to newer commits could break things unexpectedly.

However I also agree that there probably should be an option in `tar_scm` for this `--remote` behaviour, as it could be useful in some circumstances.  You could implement this simply by looking at how the existing options are implemented, and then adding a new one in a similar way.

BTW it is not necessary to close a PR and submit a new one when you notice a mistake. Also please squash both commits into a single one.  More details on both of these topics here:

- http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/

Thanks again!

nick-wang:
Thanks @adrianschroeter @aspiers for your suggestion, also thanks for link about  deal with the pull requests.
I will continue the fix in couple days later(or maybe week...) after another urgent issue.

nick-wang:
Back to fetch HEAD as default. Add an option 'master' to parameter 'submodules' to allow user fetch the latest master branch of submodules.
Also add the corresponding test case.

nick-wang:
@aspiers @adrianschroeter Could you help me to review on this? I add an parameter "master" to submodules, which enable to fetch the latest master branch of all submodules.
Fetch the detached HEAD of submodules is still the default behavior.

aspiers:
Thanks a lot, this looks very good now!  I think the only remaining thing required is to squash the two commits into a single one and then [force-push to update this pull request](http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/), then I will be happy to merge it.  Thanks again!

nick-wang:
@aspiers Thanks for your review. I squashed the two commit into one, please help to check again.

aspiers:
Great, thanks a lot!

PR #107, opened 2016-06-21
tar_scm.py: Always use an absolute path for the 'output' directory

hwoarang:
output directory is expected to be an absolute path but this is not
being enforced in the code. If a relative path is passed, it can lead
to failures since the working directoy changes during the code so
the relative path will no longer be valid.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -1155,6 +1155,7 @@ def parse_args():
     if not os.path.isdir(args.outdir):
         sys.exit("%s: No such directory" % args.outdir)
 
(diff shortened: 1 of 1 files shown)


hwoarang:
OK I just noticed this is identical to #90. So I am closing this one

hwoarang:
Re-opening per discussion in https://github.com/openSUSE/obs-service-tar_scm/pull/90

aspiers:
Thanks but please can you fix the Travis CI build?

hwoarang:
@aspiers I rebased my branch and now Travis is happy :)

PR #112, opened 2016-07-14
Allow to extract the version based on git describe --tags without shortening

DimStar77:
[2017/02/04 updated description from new improved commit message]

In order to be able to fully automatically have versions extracted from git so far one could rely on something like @PARENT_TAG@+@TAG_OFFSET@ or @PARENT_TAG@.%cd or any other sort of variation.

Those all work reasonably well, but give strange appearance for the case that the git ref being packages is exactly a tag (first would result in +0, second form still gives a commit date)

@PARENT_TAG@ is implemented by git describe --abbrev=0

This change introduces the capability to instruct the tar_scm service if abbrev should be used (=0) and retain the existing @PARENT_TAG@ behavior or use the new mode (abbrev=7, default for git describe) which in turn results in an automatic version that is "always usable":

* If the commit is a TAG ref: only shows the TAG name
* If the commit is ahead of a TAG-ref, show TAG-TAG_OFFSET-commithash

As `-` is not valid in the rpm version field, it is replaced by `.`

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -226,27 +226,36 @@ class TarSCM:
             if versionformat is None:
                 versionformat = '%ct.%h'
 
+            abbrev = args['tag_abbrev']
+            if abbrev is None:
+                abbrev = 0
+
+            version = version_iso_cleanup(safe_run(['git', 'log', '-n1',
+                                                    '--date=short',
+                                                    "--pretty=format:%s" %
+                                                    versionformat],
+                                                    repodir)[1])
+
             if not parent_tag:
-                rc, output = self.helpers.run_cmd(['git', 'describe', '--tags', '--abbrev=0'],
-                                     repodir)
+                rc, output = self.helpers.run_cmd(['git', 'describe', '--tags',
+                                                   '--abbrev={0}'.format(abbrev)], repodir)
                 if rc == 0:
                     # strip to remove newlines
                     parent_tag = output.strip()
-            if re.match('.*@PARENT_TAG@.*', versionformat):
+            if re.match('.*@PARENT_TAG@.*', version):
                 if parent_tag:
-                    versionformat = re.sub('@PARENT_TAG@', parent_tag, versionformat)
+                    version = re.sub('@PARENT_TAG@', parent_tag, version)
                 else:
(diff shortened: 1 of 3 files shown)


DimStar77:
I had this integrated on top of obs-service-tar_scm-0.6.0~pre.1461681073.57c0bab which worked as I needed it...

the rebase on top of master is causing some unexpected troubles now;

  File "/usr/lib/obs/service/tar_scm", line 541, in detect_version_git
    abbrev = args['tag_abbrev']
KeyError: 'tag_abbrev'

DimStar77:
DOH - ok, that stupid issue is also fixed...

okurz:
except the still existing typo "exract" in commit message and PR header I think this is a good idea ;-)

DimStar77:
typo in commit message fixed

DimStar77:
fixed the previous identified travis issue with python 2.6 - all tests passed

toabctl:
What about a testcase?

DimStar77:
will have to find out how to do one....

DimStar77:
now with a test-case

aspiers:
Cool, thanks! Please can you [add a proper explanation to the commit message](https://wiki.openstack.org/wiki/GitCommitMessages#Information_in_commit_messages) to help me review this?

okurz:
@aspiers DimStar77 added a lot of details to the commit message which looks very good to me. Your review and merge of this would be appreciated.

aspiers:
I will certainly review this ASAP, but I only just got back from vacation, so I'm working through the backlog. Thanks for your patience.

DimStar77:
@aspiers which numbered item on your backlog is this review by now? :)

DimStar77:
Just an occasional ping - another 2 months later

aspiers:
Also I recommend you wait for #132 to be merged before resolving conflicts, otherwise you'll have to do it twice :-/  But BTW in principle this change looks good to me so I do want it merged ASAP.  Thanks a lot for your work and patience!

aspiers:
IIUC this has been rebased in #135 so I think we can close this - if I'm wrong please let me know and I'll reopen.

Issue #111, opened 2016-07-06
need way to transform things like @PARENT_TAG@


darix:
some projects have tags which result in things like from git describe --tags:

```
v8.4.0.rc1-9121-gcfd5870
release-2.1.0-1186-g507ed2d
```

it would be nice to be able to transform the result of versionformat.

```
<services>
  <service name="obs_scm">
    <param name="url">https://gitlab.com/gitlab-org/gitlab-ce.git</param>
    <param name="versionformat">@PARENT_TAG@~git@TAG_OFFSET@.%h</param>
    <param name="revision">master</param>
    <param name="scm">git</param>
    <param name="changesgenerate">enable</param>
  </service>
[snip]
</services>
```

could be handled with something like

```
    <param name="versionformat" match='v(.*)' replace='$1'>@PARENT_TAG@~git@TAG_OFFSET@.%h</param>
```

```
    <param name="versionformat" match='release-(.*)' replace='$1'>@PARENT_TAG@~git@TAG_OFFSET@.%h</param>
```

i know executing user provided regexp can be dangerous, so i am open to better solutions.

aspiers:
Thanks - you're right, and your `match` / `replace` idea looks quite nice.  However please note that this is pretty much a duplicate of #92, and is also quite related to #60 and #55.  I'm trying to figure out a clean, general solution which makes everyone happy.

I could just merge #114 which I *think* would probably address your particular needs, since you only need to strip prefixes.  But other people need to do other regexp-based munging, and I'm not keen on implementing two layers of munging if one can do the job for everyone.  Help welcome on the design side (I expect the actual coding to be simple).  Maybe it's worth creating a wiki page which captures everyone's use cases and proposes a best-of-breed solution.

aspiers:
Note also [the security concerns](https://github.com/openSUSE/obs-service-tar_scm/issues/60#issuecomment-227417336) in #60 which would be introduced by trying to make this too flexible.

aspiers:
I've labelled all issues and PRs relating to this: https://github.com/openSUSE/obs-service-tar_scm/labels/version%20calculation

bluca:
Looks like a solution was deployed on build.opensuse.org (although cannot see it in this repo) as the following parameters are now available:

```
  <parameter name="versionrewrite-pattern">
    <description>
      Regex used to rewrite the version which is applied post versionformat. For
      example, to remove a tag prefix of "v" the regex "v(.*)" could be used.
[...]

PR #114, opened 2016-07-26
Add --stripversionprefix parameter for git tags

### LINKED ISSUE #92: v-prefixed tags in git

bluca:
Adds a new "stripversionprefix" parameter that takes a string and
strips it from the git tag starting from the left, if it exists.
Useful when using @PARENT_TAG@ and upstream prefixes "v" or "release"
to the tag, and the generated package version should not include them.

I am facing this issue so I've put together a diff following suggestions in https://github.com/openSUSE/obs-service-tar_scm/pull/100#issuecomment-231048860
Fixes #92

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -571,6 +571,11 @@ def detect_version_git(args, repodir):
 
     version = safe_run(['git', 'log', '-n1', '--date=short',
                         "--pretty=format:%s" % versionformat], repodir)[1]
+
+    if 'stripversionprefix' in args and args['stripversionprefix'] is not None:
+        if version.startswith(args['stripversionprefix']):
+            version = version[len(args['stripversionprefix']):]
+
     return version_iso_cleanup(version)
 
 
@@ -1095,6 +1100,8 @@ def parse_args():
                         help='Specify a base version as prefix.')
     parser.add_argument('--parent-tag',
                         help='Override base commit for @TAG_OFFSET@')
+    parser.add_argument('--stripversionprefix',
+                        help='Strip prefix string from git tag.')
     parser.add_argument('--revision',
                         help='Specify revision to package')
     parser.add_argument('--extract', action='append',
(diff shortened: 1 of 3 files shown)


bluca:
Not sure why Travis isn't running, but I enabled it on my fork and the same branch passes the tests:

https://travis-ci.org/bluca/obs-service-tar_scm/builds/147615621

aspiers:
Thanks!  This looks very good to me, although [as I just noted in #111](https://github.com/openSUSE/obs-service-tar_scm/issues/111#issuecomment-235599359), I'm a bit hesitant to add a feature which only strips left prefixes, when other users have requested something similar which needs more flexibility.  It would be nice if we could find a general solution which addressed all use cases.  I've labelled all related PRs and issues:

- https://github.com/openSUSE/obs-service-tar_scm/labels/version%20calculation

Thoughts very welcome on the best way to keep everyone happy at the same time!

bluca:
Sure, I understand. Unfortunately I wouldn't really know how to reconcile the security needs vs the broad general solution, given user-specified regexps are out of the question.

ddiss:
Anything holding this up, aside from the discussion around regexp support? If regexps aren't going to be supported then I'd really like to see this merged, as I think it covers the most common use cases.

aspiers:
I think regexps should be supported.  Sorry for the delay but currently I have to give more attention to #123 which is pretty urgent.

bluca:
Declining since another solution has been deployed (see https://github.com/openSUSE/obs-service-tar_scm/issues/111#issuecomment-291186606 )

PR #119, opened 2016-09-30
tar_up: fix obs URL

jirislaby:
my oscrc always had 'https://api.opensuse.org/' including the slash at
the end. It works for me, but I don't know if this is the correct fix.
If it is not, feel free to suggest a better solution :).

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -1028,11 +1028,11 @@ def get_changesauthor(args):
         return args.changesauthor
 
     config = ConfigParser.RawConfigParser()
-    obs = 'https://api.opensuse.org'
+    obs = 'https://api.opensuse.org/'
     config.add_section(obs)
     config.set(obs, 'email', DEFAULT_AUTHOR)
     config.read(os.path.expanduser('~/.oscrc'))
-    changesauthor = config.get('https://api.opensuse.org', 'email')
+    changesauthor = config.get(obs, 'email')
 
     logging.debug("AUTHOR: %s", changesauthor)
     return changesauthor


jirislaby:
On 09/30/2016, 05:36 PM, Adam Spiers wrote:
> *@aspiers* commented on this pull request.
> 
> I don't really have any objections to this, but I'm curious - did you
> ever see any negative side effects as the result of the missing trailing
> |/|?

Yes, it causes opensuse@buildservice.org to be author of the changes
instead of me.

> Also, should this really be hardcoding the OBS API endpoint? What about
> when it's used with IBS or other OBS instances?

It should handle different BS, sure.

-- 
js

aspiers:
Ohhhh, now I understand, thanks!  So this is not the correct fix, because some people (including me) actually have it the other way around in their `.oscrc`, i.e. with the slash.  So this would fix it for some people and break it for other people.  A more robust fix would try to retrieve the right email via an endpoint without the slash, and if that fails, retry with the slash.

Having said that, this would still break when invoking `tar_scm` from within a different OBS instance.  So I think that the proper fix should do the lookup based on which OBS instance is being used in the current context, although unfortunately it seems that `osc` doesn't currently pass that to the source service on invocation.  I can submit an `osc` bug for that though.

aspiers:
I submitted https://github.com/openSUSE/osc/issues/234 for this.

aspiers:
@jirislaby My fix for openSUSE/osc#234 (i.e. openSUSE/osc#235) was merged, so it should now be pretty easy to write the correct fix for this.  However we need to bear in mind this helpful hint from the `osc` maintainer about how to do it right: https://github.com/openSUSE/osc/issues/234#issuecomment-251363657

aspiers:
@jirislaby I've just filed #120 to track this issue.

PR #113, opened 2016-07-21
Adding http proxy support

doccaz:
These changes add support for using HTTP proxies with GIT, SVN, HG and BZR, according to each one's particular requirements:
- git needs to have some variables redefined for each run
- svn needs a temporary config file for each run
- hg needs some command line parameters for each run
- bzr seems to honour the http[s]_proxy variables.

I needed this to be able to use tar_scm inside a private instance of OBS behind a corporate firewall in a customer's site.

I did it by defining a "global_scm_command" variable with each particular SCM needs, and altering all safe_run() calls to use it.

PR #121, opened 2016-11-03
Add http proxy support

doccaz:
This pull request adds proper http proxy support for git/svn/bzr on the tar_scm service.

Did a few changes as suggested, including breaking up the commits between code and tests.

This pull request supersedes pull request #113 .

### DIFF
--- a/tests/bzrtests.py
+++ b/tests/bzrtests.py
@@ -2,7 +2,7 @@
 
 from commontests import CommonTests
 from bzrfixtures import BzrFixtures
-from utils       import run_bzr
+from utils import run_bzr
 
 
 class BzrTests(CommonTests):
@@ -14,9 +14,9 @@ class BzrTests(CommonTests):
     """
 
     scm = 'bzr'
-    initial_clone_command = 'bzr checkout'
-    update_cache_command  = 'bzr update'
-    sslverify_false_args  = '-Ossl.cert_reqs=None'
+    sslverify_false_args = '-Ossl.cert_reqs=None'
+    initial_clone_command = 'bzr .*checkout'
+    update_cache_command = 'bzr .*update'
     fixtures_class = BzrFixtures
 
     def default_version(self):
--- a/tests/gittests.py
+++ b/tests/gittests.py
@@ -6,10 +6,10 @@ import re
 import tarfile
 import textwrap
 
-from githgtests  import GitHgTests
+from githgtests import GitHgTests
 from gitsvntests import GitSvnTests
 from gitfixtures import GitFixtures
-from utils       import run_git
+from utils import run_git
 
 
 class GitTests(GitHgTests, GitSvnTests):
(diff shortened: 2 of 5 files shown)


doccaz:
ping?

tar_scm.py
```diff
@@ -92,62 +97,188 @@ def is_sslverify_enabled(kwargs):
     return 'sslverify' not in kwargs or kwargs['sslverify']
 
 
+def is_proxy_defined():
+    """Returns ``True`` if the ``proxy`` option has been enabled or
+    not been set (default enabled) ``False`` otherwise."""
+    if os.environ.get('http_proxy') is not None:
```
aspiers: What about if `https_proxy` is defined but not `http_proxy`?

tar_scm.py
```diff
@@ -54,6 +54,10 @@ def run_cmd(cmd, cwd, interactive=False, raisesysexit=False):
     env = os.environ.copy()
     env['LANG'] = 'C'
 
+    if is_proxy_defined():
+        env['http_proxy'] = os.environ.get('http_proxy')
```
aspiers: Is there a reason not to stick to the convention of environment variables being uppercase?

tar_scm.py
```diff
@@ -92,62 +97,188 @@ def is_sslverify_enabled(kwargs):
...
+    if os.environ.get('http_proxy') is not None:
+        return True
+    else:
+        return False
+
+
+def define_global_scm_command(scm_type, sslverify):
+    """Defines the global scm command (with or without proxy)"""
+
+    global svntmpdir
```
aspiers: No globals please - we want to keep things tightly scoped.

tar_scm.py
```diff
@@ -92,62 +97,188 @@ def is_sslverify_enabled(kwargs):
...
+
+    logging.debug('Using scm command: [' + scm_type + ']')
+    if sslverify:
+            logging.debug('SSL verification is: [ ON ]')
+    else:
+            logging.debug('SSL verification is: [ OFF ]')
+
+    # git should honor the http[s]_proxy variables, but we need to
+    # guarantee this, the variables do not work every time
+    if scm_type == 'git':
```
aspiers: This function is much too long.  Please stick to the existing code convention by splitting it up into one function for each SCM, e.g. as it's already done with `fetch_upstream_{git,hg,bzr,svn}`, plus extra functions for the code which is shared between SCMs.  You could call the SCM-specific functions `git_command`, `hg_command` etc.

In the future, ideally we could make it better by having one subclass for each SCM, derived from a base class which defines the common interface.  That would allow reuse of common code via inheritance.

tar_scm.py
```diff
@@ -92,62 +97,188 @@ def is_sslverify_enabled(kwargs):
...
+    """Defines the global scm command (with or without proxy)"""
+
+    global svntmpdir
+    global hgtmpdir
+    global global_scm_command
+    global_scm_command = ['unknown']
+
+    logging.debug('Using scm command: [' + scm_type + ']')
+    if sslverify:
+            logging.debug('SSL verification is: [ ON ]')
```
aspiers: Too much indentation here.

tar_scm.py
bmwiedemann: This variable comes in lower-case from our `/etc/profile.d/profile.sh` and is also documented as lowercase in `man curl` so should stay this way.

aspiers: OK, that's a good reason. Although it's disappointing that the `curl` and `profile.sh` authors decided to deviate from a convention widely accepted since 1979.

aspiers:
Unfortunately this will need to be rebased after the re-convergence work in #123 and #132 is complete.

aspiers:
@doccaz Would it be possible to rebase this on top of the new `master`, now that #132 is merged? Sorry to create extra work for you!

tar_scm.py
aspiers: Now that #132 is merged, we have exactly those subclasses I was wishing for.  So we can do this via a method in each SCM subclass.

doccaz:
Superseded by pull request #224 . Closing.

Issue #127, opened 2017-01-26
Support GPG signature verification


ddiss:
git commits and tags can be be GPG signed using the -S and -s parameters respectively. Verification can be performed using git verify-commit/verify-tag.

tar_scm should provide the ability to verify commit/tag GPG signatures against a public key stored in the project repository.

aspiers:
Thanks, that sounds like an excellent idea!

ddiss:
Thanks for the feedback.
I have a rough WIP set of changes in my repo at https://github.com/ddiss/obs-service-tar_scm/tree/wip_gpg_signature_check . Hoping to have them cleaned up, fully tested and submitted in the next day or so.

aspiers:
@ddiss That's awesome! Please don't hate me too much, but unfortunately you'll have to rebase and fix some merge conflicts due to a reorganisation of the code which just landed in `master` (see #125 and #128) and is still ongoing - I'm still working on merging the remainder of commits from #123.  However IIUC the main restructuring (introducing separate classes for each SCM and then splitting them out into separate files) is already in `master`.

ddiss:
Thanks! Yeah, I saw the new refactoring, and have rebased + pushed to the same branch.
It's still work-in-progress, but is starting to take shape.

aspiers:
Great.  Sorry I lied, the splitting into separate files is not there yet but will be soon.

ddiss:
Okay, is there a branch with all of the pending changes, that I can as a staging area?

aspiers:
Sorry, somehow I missed this question (for some reason GitHub wasn't sending me email for a whole bunch of notifications before).  But in any case the codebase is now stable again so it would be great if this work was rebased against latest `master`.

ddiss:
Yeah, sorry about the wait here. I've been pretty busy with other things, but I'll hopefully get back to it soon.

cryptomilk:
I hope so too :-)

bmwiedemann:
How will trusted PGP keys be specified?
Other parts of OBS use a file like zsh/zsh.keyring
Can we use the same file here or will OBS then expect a signed tarball?
or maybe we add files like
foo-commit
foo-commit.asc
foo.keyring
and leave the verification to the existing code?

ddiss:
> How will trusted PGP keys be specified?

it's currently implemented as an extra verify-revision-key tar_scm parameter

> Other parts of OBS use a file like zsh/zsh.keyring
[...]

PR #123, opened 2016-11-10
Next revision - Please Merge

M0ses:

PR #130, opened 2017-02-01
GPG signature verification

ddiss:
This patch-set implements support for git tag GPG signature verification. Although git supports GPG signing of tags and commits, these changes currently only cover use-cases where the user supplied revision directly corresponds to a signed tag.

https://github.com/openSUSE/obs-service-tar_scm/issues/127

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -1213,11 +1213,19 @@ def parse_args():
     parser.add_argument('--history-depth',
                         help='Obsolete osc service parameter that does '
                              'nothing')
+    parser.add_argument('--verify-revision-key', default='',
+                        help='Specify GPG public key to use for verification'
+			     'of the source revision')
     args = parser.parse_args()
 
     # basic argument validation
     if not os.path.isdir(args.outdir):
         sys.exit("%s: No such directory" % args.outdir)
+
+    if (args.verify_revision_key
+                      and not (os.path.isfile(args.verify_revision_key)
+                               or os.path.isabs(args.verify_revision_key))):
+        sys.exit("%s: Missing or relative key path" % args.verify_revision_key)
 
     args.outdir = os.path.abspath(args.outdir)
     orig_subdir = args.subdir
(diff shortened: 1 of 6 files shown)


ddiss:
I realise that I'll likely have to rebase this atop pending scm-backend-source-split changes, but figured it'd still be helpful to get an early review.

aspiers:
+1 on making the code available for the early review, although personally I'll be very short on time the next few weeks, so apologies if it takes a while to get to.  There are a bunch of other outstanding PRs which have also been vying for my attention.  Unfortunately #123 which was the most burning one has caused merge conflicts for all the others :-(

msmeissn:
opened an audit bug https://bugzilla.suse.com/show_bug.cgi?id=1023717

first thing to be aware is that "tag names" can contain spaces and special characters like backticks etc

ddiss:
New version includes a squashed change (https://github.com/ddiss/obs-service-tar_scm/commit/f232b789a54643451cdb838ebf51194e71101360) to escape the Git tag name and GPG key path.

ddiss:
One thing I'm still not quite sure on is the requirement that $revision corresponds to a signed tag.
Maybe it'd be better to add a new --gpg-revision-tag which is used for GPG signature verification. There's also the question of whether commit signatures should be supported in future, and if so, should they be handled via the same logic, or whether more commit-signature specific parameters should be added.
I think in it's current state it's reasonably simple, and catches the most common use case.

ddiss:
Following further feedback from the security audit, this new version:
- checks for '-' GPG key and git tag prefixes
- blocks relative GPG key paths
  - this will cause issues if the service runner doesn't resolve relative GPG key paths beforehand

aspiers:
Thanks a lot!  I'm still catching up with backlog post-FTO and -conference but I hope to get to this this month.

adrianschroeter:
you  need also to extend the .service files or you won't be able to use the parameter in OBS

ddiss:
@adrianschroeter Added, thanks.
Does the current absolute-GPG-key-path requirement make sense, or must the script also accept/resolve relative GPG key paths?

aspiers:
@ddiss Would it be possible to rebase this on top of the new `master`? Sorry to create extra work for you!

ddiss:
@aspiers No problem, will rebase when I have some time.

aspiers:
@ddiss Awesome thanks!

ddiss:
Closing this for now. Upstream has undergone many conflicting changes since my initial implementation, and I don't have time to revisit it for the time being...

PR #131, opened 2017-02-01
appimage functionality

adrianschroeter:


tar_scm.py
```diff
@@ -261,6 +261,9 @@ def detect_version(self, args, repodir):
         def get_timestamp(self, args, repodir):
             d = {"parent_tag": None, "versionformat": "%ct"}
             timestamp = self.detect_version(d, repodir)
+            # avoid future times due to broken commits
+            if datetime.datetime.fromtimestamp(timestamp) > datetime.datetime.now():
+                  timestamp = datetime.datetime.now()
             return int(timestamp)
```
M0ses: I would prefer that an exception is raised in this case. Otherwise the diff-mechanism of obscpio will detect changes on  every call of service until the broken timestamp is in the past.

M0ses: This would only work for git. Once @aspiers merged the rest of my nextRev branch, we should make this detection in TarSCM/archive.py. Then its done for all SCM types

aspiers: #132 will be merged VERY soon.

M0ses:
The appimage support of this commit is already in #132.

Our actual point of discussion is if we should prevent timestamps in the future or raise an exception and if we raise an exception, would we like to let the user overwrite with an parameter like "--use-current-timestamp".

@aspiers What is your opinion about this?

aspiers:
@M0ses I'll need a bit more context to understand the question. We can talk about it tomorrow?

M0ses:
closing this PR in favour to #132 . Functionality should be already in master. Avoiding timestamp in future issue should be discussed in #162

PR #132, opened 2017-02-04
Reconverge with OBS [part 3]

aspiers:
This is the third in a series of PRs to reconverge the codebase with the currently live code in OBS, which is represented by the changes in #123. (#125 was the first part, and #128 was the second.)

It still needs a commit to fix the PEP8 issues.


aspiers:
I think this is failing on Travis since the environment is missing both `/etc/obs/services/tar_scm` and `~/.obs/tar_scm`, so the insertion of a fake section header into the config file doesn't take effect.  So presumably

	for opt in config.options('tar_scm'):

needs to be protected with:

        if config.has_section('tar_scm'):
            for opt in config.options('tar_scm'):

although this fake section header thing makes me uncomfortable.  The `tar_scm` config file was never supposed to be in `.ini` format.

aspiers:
@M0ses The question is, why was this not failing before.  I presume that there is some new test in this PR which is now covering this code path but I haven't checked.  If so, the fix commit needs to be inserted in the sequence of commits prior to the new test, to avoid breaking `git bisect`.

PR #135, opened 2017-02-26
Allow to extract the version based on git describe --tags without shortening - rebased #112

okurz:
(Rebase of #112)

In order to be able to fully automatically have versions extracted from git so far one could rely on something like `@PARENT_TAG@+@TAG_OFFSET@` or `@PARENT_TAG@.%cd` or any other sort of variation.

Those all work reasonably well, but give strange appearance for the case that the git ref being packages is exactly a tag (first would result in +0, second form still gives a commit date)

`@PARENT_TAG@` is implemented by `git describe --abbrev=0`

This change introduces the capability to instruct the tar_scm service if abbrev should be used (=0) and retain the existing `@PARENT_TAG@` behavior or use the new mode (abbrev=7, default for git describe) which in turn results in an automatic version that is 'always usable':

* If the commit is a TAG ref: only shows the TAG name
* If the commit is ahead of a TAG-ref, show TAG-TAG_OFFSET-commithash

Since `-` is not valid in the rpm version field, it is replaced by `.`.

### DIFF
--- a/tar_scm.py
+++ b/tar_scm.py
@@ -226,27 +226,31 @@ class TarSCM:
             if versionformat is None:
                 versionformat = '%ct.%h'
 
+            abbrev = args['tag_abbrev'] if args['tag_abbrev'] else None
+
+            version = version_iso_cleanup(self.helpers.safe_run(['git', 'log', '-n1', '--date=short',
+                    "--pretty=format:%s" % versionformat], repodir)[1])
+
             if not parent_tag:
-                rc, output = self.helpers.run_cmd(['git', 'describe', '--tags', '--abbrev=0'],
-                                     repodir)
+                rc, output = self.helpers.run_cmd(['git', 'describe', '--tags',
+                                                   '--abbrev={0}'.format(abbrev)], repodir)
                 if rc == 0:
                     # strip to remove newlines
                     parent_tag = output.strip()
-            if re.match('.*@PARENT_TAG@.*', versionformat):
+            if re.match('.*@PARENT_TAG@.*', version):
                 if parent_tag:
-                    versionformat = re.sub('@PARENT_TAG@', parent_tag, versionformat)
+                    versionformat = re.sub('@PARENT_TAG@', parent_tag, version)
                 else:
                     sys.exit("\033[31mNo parent tag present for the checked out "
                              "revision, thus @PARENT_TAG@ cannot be expanded.\033[0m")
 
-            if re.match('.*@TAG_OFFSET@.*', versionformat):
+            if re.match('.*@TAG_OFFSET@.*', version):
                 if parent_tag:
                     rc, output = self.helpers.run_cmd(['git', 'rev-list', '--count',
                                           parent_tag + '..HEAD'], repodir)
                     if not rc:
                         tag_offset = output.strip()
-                        versionformat = re.sub('@TAG_OFFSET@', tag_offset,
-                                               versionformat)
+                        version = re.sub('@TAG_OFFSET@', tag_offset, version)
                     else:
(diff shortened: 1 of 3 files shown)


okurz:
I would invest the time to rebase and fix but I am reluctant to do so when it will just hang around so I don't see the point.

DimStar77:
@okruz I understand your sentiment - my original submission for this was in Jul 2016 - and kept on being 'reviewed soon'; even though I still think it would be a nice addition to have

DimStar77:
ping - is there REALLY no interest in this? The original submission was done 1.5 years ago

aspiers:
It's nothing to do with lack of interest, and everything to do with lack of time - being co-maintainer of this repo is virtually nothing to do with my day job any more and just something I do because I want to contribute to openSUSE.  @adrianschroeter and @M0ses are also maintainers, and I suspect all 3 of us would be very open to the idea of other people stepping up as co-maintainers too.  We all acknowledge that development isn't moving at an ideal pace, but in the FL/OSS world that's hardly unusual and can only be fixed if other people are willing to step up.  I'm very happy to have a conversation about the future of the project and maintainership - if anyone wants that, please submit a GitHub issue for it and we can discuss there.

On this specific issue, it's complicated because there are multiple PRs and issues all in the same area (I've labelled them `version calculation`) and there are various degrees of overlap and conflict between them.  The delay here is because it needs someone to look at all of them holistically and figure out the best way forward, and until today I didn't personally make a start on that, nor did anyone else volunteer (admittedly I could have been more explicit about requesting help - mea culpa - but equally OTOH noone asked "hey are you blocked on this somehow? do you need help?", which I would have been far more receptive to).

Having said all that, I have started to rebase this PR, and in the process am gaining a better understanding of all of these related issues. Without help from anyone else I won't promise anything by a certain date, but am hoping to make further progress over the next few days (and again this is nothing to do with my day job, which is imposing some very urgent work at the moment).

tar_scm.py
```diff
@@ -799,12 +801,12 @@ def cleanup(dirs):
         shutil.rmtree(d)
 
 
-def version_iso_cleanup(version):
+def version_iso_cleanup(version, char=''):
```
M0ses: could you please explain, why an additional parameter is added here as I cannot see where the default (char='') would be changed.

tar_scm.py
```diff
@@ -226,40 +226,42 @@ def detect_version(self, args, repodir):
...
             if not parent_tag:
-                rc, output = self.helpers.run_cmd(['git', 'describe', '--tags', '--abbrev=0'],
-                                     repodir)
+                rc, output = self.helpers.run_cmd(['git', 'describe', '--tags',
+                                                   '--abbrev={0}'.format(abbrev)], repodir)
                 if rc == 0:
                     # strip to remove newlines
                     parent_tag = output.strip()
-            if re.match('.*@PARENT_TAG@.*', versionformat):
+            if re.match('.*@PARENT_TAG@.*', version):
```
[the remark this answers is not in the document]
okurz: I don't remember. Maybe @DimStar77 knows as he edited this section

tar_scm.py
okurz: It's used as a fake static function variable

DimStar77: in the original PR (which okurz kindly rebased into this one) there was at least one point where ```char='.'``` was passed, thus making this change necessary.

tar_scm.py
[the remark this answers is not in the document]
DimStar77: I see how this can be breaking compared to what we had before :(

okurz:
this is getting too old …

PR #136, opened 2017-02-27
Alternative "Reconverge with OBS [part 3]" #132

okurz:


okurz:
includes one additional commit to #132: "Adapt to current PEP8 styles"

aspiers:
Hey @okurz thanks a lot for your help on this!  I was deliberately holding off on the PEP8 fixes until the reconvergence reaches the point in @M0ses's branch that we're converging with where he fixes the PEP8 issues.  The reason is that if we apply the fixes before then, it is likely to cause a load of merge conflicts and makes the remainder of the reconvergence a lot harder.  At least that's what happened with the earlier stages of reconvergence, although admittedly I haven't checked with these later stages yet.

okurz:
On Wednesday, 1 March 2017 07:33:47 CET Adam Spiers wrote:
> Hey @okurz thanks a lot for your help on this!  I was deliberately holding
> off on the PEP8 fixes until the reconvergence reaches the point in @M0ses's
> branch that we're converging with where he fixes the PEP8 issues.  The
> reason is that if we apply the fixes before then, it is likely to cause a
> load of merge conflicts and makes the remainder of the reconvergence a lot
> harder.  At least that's what happened with the earlier stages of
> reconvergence, although admittedly I haven't checked with these later
> stages yet.

No worries. It was merely an experiment on my side and by doing some style 
related changes I can learn the code :-) I would be ok to redo the job after 
you could merge dependencies first.

aspiers:
I fixed all the issues in #132 and it was just merged.

Issue #137, opened 2017-03-02
Support %ci, %h and @GIT_OFFSET@ in .obsinfo


lbeltrame:
As discussed with AdrianS on #opensuse-buildservice, this would help projects creating their own services to more accurately set versions from these information.

lbeltrame:
I can try to make a stab at doing this myself, however, I'm thinking about possible approaches:

1. Duplicate what `detect_version` is doing (not my favorite idea)
2. Call `detect_version` when obsinfo is being generated (might be overkill)

PR #138, opened 2017-03-02
Add support for %ci, %h and @TAG_OFFSET@ to obsinfo files

lbeltrame:
This is done by repeated calls to get_version, so this part may be up
for optimizations.

Tests still pass (except bzr, but because it's broken locally for me).

This should fix issue #137. Notice: I didn't find anything in the test suite for testing `obsinfo` files.


lbeltrame:
Rebased on top of latest master. Please review when you can.

aspiers:
Oh, I just noticed you said there are no tests for obsinfo yet. That's not good ... Well, this is a good excuse to add them ;-)

lbeltrame:
Good point about tests. I'll see if I can do something about it. Where is a good point to check for existing tests, and where should they best put in?

aspiers:
Firstly, you should read [`TESTING.md`](https://github.com/openSUSE/obs-service-tar_scm/blob/master/TESTING.md) if you didn't already.

Next, it'll be a question of adding tests under the `tests/` subdirectory.  Unfortunately due to the limited capacity of my brain, I can't remember anything about obsinfo, so until #155 is resolved I can't make any more specific suggestions.  But let me know if you have questions about the way the existing tests work.

aspiers:
@lbeltrame I think we have at least one obsinfo test now, thanks to all the great work from @M0ses:

https://github.com/openSUSE/obs-service-tar_scm/blob/42fa499ea9026c9e03be4e7d3d9030258ea0eff1/tests/unittestcases.py#L208

However I'm not sure exactly what that test is supposed to be testing, and currently it's broken anyway.  Maybe you can liase with @M0ses to figure out the best way to write the test for the new functionality in this PR?

TarSCM/archive.py
```diff
@@ -104,6 +104,15 @@ def create_archive(self, scm_object, **kwargs):
...
             metafile.write("commit: " + commit + "\n")
+            names = {"%ci": "commit_date: ",
+                     "%h": "short_commit_hash: ",
+                     "@TAG_OFFSET@": "commits_since_last_tag: "}
+
+            for field in ("%h", "%ci", "@TAG_OFFSET@"):
+                fake_args = args.copy()
+                fake_args["versionformat"] = field
+                value = str(scm_object.detect_version(fake_args))
+                metafile.write(names[field] + value + "\n")
```
M0ses: looks very specific to git to me. Maybe this should be handled in a SCM specific method and TarSCM.scm.base should have a generic method which simply 'pass'. Otherwise I guess it could break e.g. svn/bzr/...

M0ses:
I think, we need at least some unittests for each SCM here, but maybe some integration test would be better.
@aspiers - you mentioned one of my test cases, which is broken.
Anyway, I think this would have not prevented us from breaking other SCM's. 
Maybe we could consider this in https://github.com/openSUSE/obs-service-tar_scm/issues/148

TarSCM/archive.py
aspiers: Yeah I agree, git-specific stuff should remain in the `Git` class.

PR #149, opened 2017-04-13
add official spec file to git

adrianschroeter:


adrianschroeter:
will use this in the packages using extract feature once merged.

obs-service-tar_scm.spec
```diff
@@ -0,0 +1,166 @@
...
+BuildRequires:  mercurial
+BuildRequires:  python >= 2.6
+%if 0%{?fedora_version} || 0%{?rhel_version} || 0%{?centos_version}
+BuildRequires:  PyYAML
+%else
+BuildRequires:  python-PyYAML
+%endif
+BuildRequires:  python-dateutil
+BuildRequires:  python-lxml
+BuildRequires:  python-mock
```
aspiers: I think we also need `python-unittest2` since #132 started using it.

M0ses: I doubt, because it is not in the spec file of [openSUSE:Tools](https://build.opensuse.org/package/view_file/openSUSE:Tools/obs-service-tar_scm/obs-service-tar_scm.spec?expand=1) but this package  builds.

Please merge and lets fix problems afterwards

adrianschroeter: I am a bit confused here, we run the testsuite during package build, but it seems we don't miss the dependency.

So, either we don't run all test cases or this isn't needed?

aspiers: `unittest2` is required for Python < 2.7 only.

M0ses:
Hmm, seems like it is outdated - https://build.opensuse.org/package/view_file/openSUSE:Tools/obs-service-tar_scm/obs-service-tar_scm.spec?expand=1 should be the most recent version

M0ses:
Ok, as already discussed with adrian - I will take over here, updating the spec file from OBS and will have a look at the unittest2 deps.

PR will follow after
https://build.opensuse.org/package/show/openSUSE:Tools/obs-service-tar_scm
is ready

Issue #150, opened 2017-04-13
need better linting


aspiers:
Currently we only have PEP8.  `flake8` finds other issues, and `pylint` is even more comprehensive.  I've started [a `linting` branch](https://github.com/aspiers/obs-service-tar_scm/compare/reconverge...aspiers:linting) to harness these.  The "only" work remaining is to fix all the `pylint` issues, but there are loads of them.

delirious-lettuce:
@aspiers ,

I could try and fix some `pylint` issues but I'm not sure if I should start before #151 is merged?

aspiers:
@delirious-lettuce On the contrary, we cannot merge #151 until it addresses all the `pylint` issues, since otherwise Travis will fail on the PR.  So your help in finishing #151 would be very welcome!  You could simply start with that branch and add a commit for each category of `pylint` failure, for instance.

delirious-lettuce:
@aspiers ,

Ah, of course. That makes sense. Looking back at my comment, I think I was confused because there wasn't a linting branch on this repo. I guess I should just fork your fork (where the linting branch is) and push the changes there, right?

aspiers:
No need to send PRs to my PR :-)  Just start with my branch as the base and then do whatever you want to it and push a PR upstream which supersedes #151.

delirious-lettuce:
@aspiers ,

Sorry for the confusion! That's exactly what I'll do, thanks!

M0ses:
Hmm, personally I would prefer if we could cleanup master first step by step and then activate pylint/flake8.
This way we could distribute the work to more people and could implement new features in the meanwhile. Each module which gets touched while implementing an new bugfix/feature should be sanitized in separate commits inside the PR.

So hopefully we could avoid such a big divergence like we had the last time.

@aspiers What do you think about this?

aspiers:
Thanks @M0ses, that's a much better idea!

M0ses:
Some additional thoughts:

The tests directory should not be checked or at least should be checked with a different pylintrc.
Problems should be only reported, because we would like / should support different version of pylint (2.6?/2.7/3.x)

For example, these warnings of pyline-3.*

************* Module tests.commontests
W: 39, 8: Using deprecated method assertEquals() (deprecated-method)
W: 51, 8: Using deprecated method assertRegexpMatches() (deprecated-method)
W: 66, 8: Using deprecated method assertRegexpMatches() (deprecated-method)
[...]

PR #151, opened 2017-04-13
Improve linting (#150)

aspiers:
This starts addressing #150, but we cannot hook up the new linters to Travis until the very many pylint issues are fixed.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -37,7 +37,10 @@ PY_FILES=$(shell $(LIST_PY_FILES))
 default: check
 
 .PHONY: check
-check: pep8 test
+check: lint test
+
+.PHONY: lint
+lint: pep8 flake8
 
 .PHONY: list-py-files
 list-py-files:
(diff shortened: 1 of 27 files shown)


GNUmakefile
```diff
@@ -37,7 +37,10 @@ PY_FILES=$(shell $(LIST_PY_FILES))
 default: check
 
 .PHONY: check
-check: pep8 test
+check: lint test
+
+.PHONY: lint
+lint: pep8 flake8
```
Ana06: `flake8` also verifies `pep8`, right? so you are running pep8 twice. :thinking:

GNUmakefile
```diff
@@ -49,7 +52,23 @@ pep8: $(PY_FILES)
...
+	fi
+	$(LIST_PY_FILES) | xargs flake8 --ignore=E221,E241,E251,E272,E731
+
+.PHONY: pylint
+pylint: $(PY_FILES)
+	@if ! which pylint >/dev/null 2>&1; then \
+		echo "pylint not installed!  Aborting." >&2; \
+		exit 1; \
+	fi
+	$(LIST_PY_FILES) | xargs pylint
```
Ana06: what about using `-j` option for pylint? It should be faster. :thinking:

GNUmakefile
Ana06: and why not adding `pylint` here? :confused:

TarSCM/__init__.py
```diff
@@ -31,8 +28,8 @@ def run():
 
     try:
         task_list.process_list()
-    except OptionsError as e:
-        print(e)
+    except OptionsError as exc:
```
Ana06: why this change? :confused:

requirements.txt
```diff
@@ -1,6 +1,8 @@
+flake8
 lxml
 mock
 pep8
```
Ana06: do you still want to run pep8 alone? if not this can also be removed, and some code in the GNUmakefile too.

TarSCM/__init__.py
M0ses: Because pylint otherwise is complaining about invalid-name. pylint wants variable names to be at least 3 char long.

M0ses:
could we close this PR in favor to https://github.com/openSUSE/obs-service-tar_scm/pull/171?

M0ses:
oops, closed accidently

aspiers:
Obsoleted by #171 and maybe others.

GNUmakefile
M0ses: Thanks for your hint but it did not make any difference in my tests on my laptop (8 cores - -j 8). Took always about 5sec.

PR #152, opened 2017-04-13
skip broken tests temporarily

M0ses:
This test cases is broken while rpm building because
it relies on a .git directory which is excluded while packaging the
sources. Test case needs refactoring to use the GitFixtures class


aspiers:
@M0ses I'm looking back at this and wondering why Travis CI succeeded in #132 before this patch - can you explain that?

PR #159, opened 2017-05-29
Fix empty appimage

M0ses:
* do not break if build section doesn`t exists in yml
* various changes for pylint

### DIFF
--- a/TarSCM/tasks.py
+++ b/TarSCM/tasks.py
@@ -29,15 +29,16 @@ class tasks():
         self.helpers        = helpers()
         self.changes        = changes()
         self.scm_object     = None
+        self.data_map       = None
 
     def cleanup(self):
         """Cleaning temporary directories."""
         logging.debug("Cleaning: %s", ' '.join(self.cleanup_dirs))
 
-        for d in self.cleanup_dirs:
-            if not os.path.exists(d):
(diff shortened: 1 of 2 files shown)


TarSCM/tasks.py
```diff
@@ -119,7 +126,6 @@ def _process_single_task(self, args):
         # self.scm_object is need to unlock cache in cleanup
         # if exception occurs
         self.scm_object = scm_object   = scm_class(args, self)
-        helpers      = scm_object.helpers
```
aspiers: Technically this removal should be in a separate commit.

aspiers:
Just expanding a bit on my request for greater clarity:

- The PR description says "do not break if build section doesn`t exists in yml"
- The corresponding commit message says "fix for empty appimage.yml"

So it's not clear to me whether this change is targeting a missing appimage.yml file, or a missing build section within a .yml file.  This is probably because of my ignorance in this area: I've never worked with AppImages at all, let alone building AppImages in OBS.  Hopefully that serves as an example of why a more comprehensive commit message as described in that presentation by @cmurphy is worth the extra 30 seconds to write.

M0ses:
@aspiers Thanks for your feedback - I tried to improve #160 with your suggestions. I will close this PR in favour to #160

Issue #147, opened 2017-04-12
slow tests should be skippable


aspiers:
Even if we do #146 there are bound to be some tests which are unavoidably slow.  For example `svn` has terrible performance :-)  So we need a mechanism for skipping the slow tests.  This should be a simple matter of writing [a decorator to skip slow tests](https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures).

M0ses:
What is the definition of "slow"? 
Would you like to track it down to test cases (functions or methods) or classes(e.g. HgTests)?

I want to help but i do not understand the goal fully.

I did a trace which test classes take most of the time:

```
########## UnitTestCases
.............ss....s..
----------------------------------------------------------------------
Ran 22 tests in 0.078s

OK (skipped=3)
0.37user 0.06system 0:00.47elapsed 91%CPU (0avgtext+0avgdata 27316maxresident)k
4640inputs+0outputs (1major+10419minor)pagefaults 0swaps
########## TasksTestCases
.........
----------------------------------------------------------------------
Ran 9 tests in 0.047s

OK
0.32user 0.03system 0:00.36elapsed 98%CPU (0avgtext+0avgdata 26024maxresident)k
232inputs+8outputs (0major+5434minor)pagefaults 0swaps
########## SCMBaseTestCases
.
----------------------------------------------------------------------
Ran 1 test in 0.004s

OK
0.26user 0.03system 0:00.30elapsed 99%CPU (0avgtext+0avgdata 25680maxresident)k
8inputs+0outputs (0major+5368minor)pagefaults 0swaps
########## GitTests
.........................................................
----------------------------------------------------------------------
Ran 57 tests in 26.869s

OK
15.70user 11.71system 0:27.34elapsed 100%CPU (0avgtext+0avgdata 34408maxresident)k
21025inputs+102040outputs (12major+3756229minor)pagefaults 0swaps
########## SvnTests
..........................................
----------------------------------------------------------------------
Ran 42 tests in 66.416s

OK
31.56user 12.04system 1:06.85elapsed 65%CPU (0avgtext+0avgdata 41508maxresident)k
12568inputs+247448outputs (27major+3559449minor)pagefaults 0swaps
########## HgTests
..................................
----------------------------------------------------------------------
Ran 34 tests in 76.240s

OK
63.75user 11.45system 1:16.60elapsed 98%CPU (0avgtext+0avgdata 29988maxresident)k
[...]

PR #161, opened 2017-05-30
[tests] new env var TAR_SCM_TC to select test cases

### LINKED ISSUE #146: test suite needs optimising
It's currently too hard to get quick feedback from the test suite, as there are too many tests taking too long.  We should do some performance profiling on the tests and tackle the bottlenecks.

M0ses:
Without this patch you have to edit the tests/test.py and comment out
tests you dont want to run in an array called testclasses.

This patch gives you the possibilty to run only selected testclasses
by specifing an environment variable with a comma separated list of
testclasses to run, e.g.

TAR_SCM_TC=UnitTestCases,TasksTestCases PYTHONPATH=. tests/test.py

This is only the first step in direction to fix #146 and #147

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -22,11 +22,17 @@ if sys.version_info < (2, 7):
 else:
     import unittest
 
+
+def str_to_class(string):
+    '''Convert string into class'''
+    return getattr(sys.modules[__name__], string)
+
 if __name__ == '__main__':
     suite = unittest.TestSuite()
     testclasses = [
         # If you are only interested in a particular VCS, you can
-        # temporarily comment out any of these:
+        # temporarily comment out any of these or use the env variable
+        # TAR_SCM_TC=<comma_separated_list> test.py
         UnitTestCases,
         TasksTestCases,
         SCMBaseTestCases,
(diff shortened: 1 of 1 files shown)


M0ses:
In this commit I splitted up each single pylint fix into a commit, but this is very inefficient as I`m writing more commit messages than code. Could we agree to do all the lint fixes in one commit (including the pylint report in the commit message)?

aspiers:
I would prefer to keep them separate otherwise there are way too many different types of change to review.  However I think this is one of the rare occasions where a detailed commit message is not needed - as long as the commit message title makes it clear which category of pylint check is being fixed, that should be sufficient.  So the overhead should not be big.  A way to reduce it even further is to follow your good suggestion of submitting a commit per category of pylint check which fixes issues across the whole codebase for that category.

M0ses:
Ok, what do you think about the change despite the lint changes? I hate to edit the tests/test.py to get rid of the other tests, especial because bzr based test cases are broken permanently as bzr is broken in TW ATM

aspiers:
I'm in Israel but back Wed. Should hopefully get time to review this in next day or 3.

tests/test.py
```diff
@@ -29,7 +29,7 @@ def str_to_class(string):
     return getattr(sys.modules[__name__], string)
 
 if __name__ == '__main__':
-    suite = unittest.TestSuite()
+    SUITE = unittest.TestSuite()
```
M0ses: Even if i isolate the into two functions (prepare_testclasses and prepare_testsuite) the problem with the constants persists. I don`t care if we disable the pylint checks as suggested in the link you posted (solution 2) or if we simply write the variable name in upper case letters.

the problem persists for

* suite 
* runner
* result

PR #168, opened 2017-06-18
Sort tar file list

bmwiedemann:
Sort tar file list
to generate tar files in a reproducible way.
    
See https://reproducible-builds.org/ for why this matters.
    
In addition to this change,
to generate completely reproducible tarballs,
the build system also needs
https://github.com/python/cpython/pull/2263


aspiers:
@bmwiedemann Thanks, now I think I understand. So to clarify, the CPython PR does not need to be merged before this one is merged, right?  You were just mentioning that as an FYI relating to the reproducible build initiative as a whole?

bmwiedemann:
yes. They are independent, but to generate reproducible tarballs in the end, it needs both of them.

Issue #145, opened 2017-04-12
need to determine support / test matrices


aspiers:
We need to be clearer about which OS+Python+SCM stacks we support.  Any supported stack should have CI, but additionally we should have CI for stacks which we want to add support in the future, e.g. Python 3.5.

## Python versions

Currently Travis tests against Python 2.6 and 2.7 only.  We need to start testing against Python 3.x, which requires deciding exactly which 3.x versions to support, and also ideally we could drop Python 2.6 which is ancient.

## Distributions

Currently [OBS tries to build `obs-service-tar_scm` against a huge number of distros](https://build.opensuse.org/package/show/openSUSE:Tools/obs-service-tar_scm), but not all of them succeed.  Which ones do we need to work?

## SCMs

Does anyone actually use `tar_scm` with `bzr`?

aspiers:
Apparently we still need 2.6 for SLE11 :-/

Good news regarding 3.x: [you can add rows to the Travis CI test matrix which are allowed to fail](https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail).

M0ses:
Required in follow up PR to https://github.com/openSUSE/obs-service-tar_scm/pull/149

M0ses:
@aspiers @adrianschroeter Do we really need 2.6 (SLE11)? AFAICS obs-service-tar_scm is not maintained in any product and OBS support for sle11 was also dropped previously. On the other hand, fixing issue/150 gets nearly impossible when we want to keep support for python 2.6. flake8/pylint/astroid (and maybe others) need special versions to run in python 2.6.

adrianschroeter:
I like to keep support SLE 11, esp. since we make use of these service also at build time.

But there is no need to run the lint checks on SLE 11 / python 2.6. Focusing on one platform should be really enough here.

aspiers:
@adrianschroeter Totally agree!

M0ses:
@adrianschroeter Do I see this correctly that we have to support potentially every single distribution, that someone could choose in OBS?

PR #171, opened 2017-06-24
Pylint class cleanup

M0ses:


aspiers:
@M0ses Looks promising! Just checking you saw that Travis failed?

adrianschroeter:
as written in https://github.com/openSUSE/obs-service-tar_scm/issues/145 I think we can disable lint errors for python 2.6. It should be enough to do style checks on one platform.

M0ses:
@aspiers: Yes, I saw these errors and tried to fix them, but I have to agree with @adrianschroeter, we should concentrate on one python version with linting - otherwise it gets a hassle - so, if you agree I would disable linting if python version < 2.7.

Then I would like to merge this PR because it is getting much too big and I would like to avoid mixing to much things in one PR

aspiers:
@M0ses I agree 100%, we only need to lint on one Python version. Once this is passing tests then we can merge straight away. (Please ensure that the commit which includes linting in the Travis run comes last, so that there are no commits which would fail the test suite.) Thanks!

M0ses:
@aspiers I reorder some commits, so now we should not have any breakage of tests in the whole PR. I didn`t test each step - but the additional linting tests are now activated with the last commit. I think we are ready to merge now.

TarSCM/scm/git.py
```diff
@@ -122,42 +123,17 @@ def detect_version(self, args):
...
-                if args['match_tag']:
-                    cmd.append("--match=%s" % args['match_tag'])
-            except KeyError:
-                pass
-            rc, output = self.helpers.run_cmd(cmd, self.clone_dir)
-
-            if rc == 0:
-                # strip to remove newlines
-                parent_tag = output.strip()
+            parent_tag = self._detect_parent_tag(args)
```
aspiers: This commit seems to have some refactorings or other parent_tag related changes which are unrelated to linting cleanups - is that deliberate?  Not sure what those changes are for, so I can't really review them.

M0ses: Sorry for confusion - Yeah it is some kind of refactoring but it was requested by pylint

`R:115, 4: Too many branches (13/12) (too-many-branches)`

because even it may be confusing, it is consistent in terms of making this modules lint-ready.

M0ses:
Thanks for reviewing in the last minutes before your vacation. I will merge it as is, as I want to make a new release next week. Have a nice vacation!

TarSCM/scm/git.py
aspiers: OK cool thanks for explanation :)

GNUmakefile
```diff
@@ -34,22 +141,41 @@ mycfgdir = $(SYSCFG)/obs/services
...
+
+PYLINT2 = $(call first_in_path_opt,$(ALL_PYLINT2))
+PYLINT3 = $(call first_in_path_opt,$(ALL_PYLINT3))
+
 default: check
 
-.PHONY: check
-check: pep8 test
+.PHONY: check check_all
+check: pep8 flake8 pylint test # test3
```
Ana06: `flake8` also verifies `pep8`, right? so you are running `pep8` twice. :thinking:

PR #172, opened 2017-07-01
[dist] changed to %py_compile in spec file

M0ses:
Without this patch, rpmlint complains about

obs-service-obs_scm-common.noarch: W: python-bytecode-inconsistent-mtime /.../TarSCM/cli.pyc 2017-06-19T11:14:47 /.../TarSCM/cli.py 2017-07-01T06:59:23

This patch disables "compile" as dep for "install" in Makefile and uses
"%py_compile" in the "%build" section of the spec file

tested for Tumbleweed with local osc build


M0ses:
@adrianschroeter Please Review

Issue #175, opened 2017-07-04
.git does not get stripped off from archive name


M0ses:
after renaming scm classes to camel case, scm extensions gets no longer stripped off from the repo base name:

```
[   99s] Running build time source services...
[  100s] Compressed obs-testpackage.git-20170425.tar to obs-testpackage.git-20170425.tar.xz
[  100s] /usr/lib/obs/service/set_version:32: RuntimeWarning: install 'packaging' to improve python package versions
[  100s]   RuntimeWarning)
[  101s] -----------------------------------------------------------------
[  101s] I have the following modifications for obs-testpackage.spec:
[  101s] 20c20
[  101s] < Release:	0
[  101s] ---
[  101s] > Release:	1.1
[  101s] -----------------------------------------------------------------
[  101s] ----- building obs-testpackage.spec (user abuild)
[  101s] -----------------------------------------------------------------
[  101s] -----------------------------------------------------------------
[  101s] + exec rpmbuild -ba --define '_srcdefattr (-,root,root)' --nosignature --define 'disturl obs://private/home:Admin/openSUSE_Leap_42.2/4a5e632913df46842b8245ec14b3fa82-obs-testpackage' /home/abuild/rpmbuild/SOURCES/obs-testpackage.spec
[  101s] error: File /home/abuild/rpmbuild/SOURCES/obs-testpackage-20170425.tar.xz: No such file or directory
```

PR #176, opened 2017-07-04
[bugfix] Decoupled self.scm from class name in TarSCM/scm/*

M0ses:
* fixes https://github.com/openSUSE/obs-service-tar_scm/issues/175

Without this patch the class name is used as self.scm in TarSCM.scm.*
classes. This value is also used to strip off the scm extension from the
url to get the repodir

Example:

https://github.com/openSUSE/obs-service-tar_scm.git -> obs-service-tar_scm

This broke after renaming scm classes to camel case, for better linting
results.

In this patch, self.scm is changed to an class attribute
and decoupled from the class name.


adrianschroeter:
approved as discussed on the phone.

PR #179, opened 2017-07-20
Debian git workflow: add obs_gbp to support git-buildpackage

bluca:
Debian + git developers workflow never include the .dsc file in the
repository, as it is the output of a build, not the input.
This makes it awkward and convoluted to use tar_scm with Debian git
repositories, and just unfeasable with obs_scm.

The most common workflow for Debian developers involves a tool called
git-buildpackage, that can generate the Debian source artefacts
(dsc file, orig.tar.gz and if non-native debian.tar.gz or diff.gz)
from a git repository.

Add a new class in the archive module that calls git-buildpackage as
an alternative to creating a tar archive or an obscpio.

This makes it possible to take advantage of the shared git clones
cache that obs_scm uses, saving a lot of network I/O.

The code is adapted from obs-service-git_buildpackage:
https://github.com/jblunck/obs-service-git_buildpackage

Example of a _service:

```
<services>
  <service name="obs_gbp">
    <param name="url">git://github.com/bluca/czmq.git</param>
    <param name="scm">git</param>
    <param name="revision">obs_gbp</param>
    <param name="versionformat">@PARENT_TAG@+git%cd</param>
    <param name="versionrewrite-pattern">(v|debian/)(.*)</param>
    <param name="versionrewrite-replacement">\2</param>
    <param name="gbp-dch-release-update">enable</param>
    <param name="extract">packaging/redhat/czmq.spec</param>
  </service>
  <service name="set_version" mode="buildtime">
    <param name="file">czmq.spec</param>
  </service>
</services>
```

---
This change is [Reviewable](https://reviewable.io/reviews/opensuse/obs-service-tar_scm/179)

### DIFF
--- a/TarSCM/cli.py
+++ b/TarSCM/cli.py
@@ -144,8 +144,14 @@ class Cli():
         parser.add_argument('--history-depth',
                             help='Obsolete osc service parameter that does '
                                  'nothing')
-        # This option is only used in test cases, in real life you would call
-        # obs_scm instead
+        parser.add_argument('--gbp-build-args', type=str,
+                            default='-nc -uc -us -S',
+                            help='Parameters passed to git-buildpackage')
+        parser.add_argument('--gbp-dch-release-update',
+                            choices=['enable', 'disable'], default='disable',
+                            help='Append OBS release number')
+        # These option is only used in test cases, in real life you would call
+        # obs_scm or obs_gbp instead
         parser.add_argument('--use-obs-scm', default = False,
                             help='use obs scm (obscpio) ')
 
@@ -158,6 +164,8 @@ class Cli():
                                  ' Set locale while service run')
         parser.add_argument('--encoding',
                             help='set encoding while service run')
+        parser.add_argument('--use-obs-gbp', default = False,
+                            help='use obs gbp (requires git-buildpackage) ')
 
         self.verify_args(parser.parse_args(options))
 
@@ -191,6 +199,9 @@ class Cli():
         args.package_meta    = bool(args.package_meta == 'yes')
         args.sslverify       = bool(args.sslverify != 'disable')
         args.use_obs_scm     = bool(args.use_obs_scm)
+        args.use_obs_gbp     = bool(args.use_obs_gbp)
+        args.gbp_dch_release_update = bool(args.gbp_dch_release_update
+                                           != 'disable')
 
         # Allow forcing verbose mode from the environment; this
         # allows debugging when running "osc service disabledrun" etc.
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -171,6 +171,7 @@ class Git(Scm):
         """
         parent_tag = args['parent_tag']
         versionformat = args['versionformat']
+        debian = args.get('use_obs_gbp', False)
         if versionformat is None:
             versionformat = '%ct.%h'
 
@@ -192,7 +193,7 @@ class Git(Scm):
                                    "--pretty=format:%s" % versionformat],
             self.clone_dir
         )[1]
-        return self.version_iso_cleanup(version)
+        return self.version_iso_cleanup(version, debian)
 
     def _detect_parent_tag(self, args):
         parent_tag = ''
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -177,6 +177,8 @@ install: dirs tar_scm service compile
 	ln -s tar_scm $(DESTDIR)$(mylibdir)/appimage
 	[ ! -L $(DESTDIR)$(mylibdir)/snapcraft ] || rm $(DESTDIR)$(mylibdir)/snapcraft
 	ln -s tar_scm $(DESTDIR)$(mylibdir)/snapcraft
+	[ ! -L $(DESTDIR)$(mylibdir)/obs_gbp ] || rm $(DESTDIR)$(mylibdir)/obs_gbp
+	ln -s tar_scm $(DESTDIR)$(mylibdir)/obs_gbp
 	find ./TarSCM/ -name '*.py*' -exec install -D -m 644 {} $(DESTDIR)$(mylibdir)/{} \;
 
 .PHONY: dirs
@@ -191,8 +193,9 @@ service: dirs
 	install -m 0644 tar.service $(DESTDIR)$(mylibdir)/
 	install -m 0644 snapcraft.service $(DESTDIR)$(mylibdir)/
 	install -m 0644 appimage.service $(DESTDIR)$(mylibdir)/
-	sed -e '/^===OBS_ONLY/,/^===/d' -e '/^===/d' tar_scm.service.in > $(DESTDIR)$(mylibdir)/tar_scm.service
-	sed -e '/^===TAR_ONLY/,/^===/d' -e '/^===/d' tar_scm.service.in > $(DESTDIR)$(mylibdir)/obs_scm.service
+	sed -e '/^===OBS_ONLY/,/^===/d' -e '/^===GBP_ONLY/,/^===/d' -e '/^===/d' tar_scm.service.in > $(DESTDIR)$(mylibdir)/tar_scm.service
+	sed -e '/^===TAR_ONLY/,/^===/d' -e '/^===GBP_ONLY/,/^===/d' -e '/^===/d' tar_scm.service.in > $(DESTDIR)$(mylibdir)/obs_scm.service
+	sed -e '/^===OBS_ONLY/,/^===/d' -e '/^===TAR_ONLY/,/^===/d' -e '/^===/d' tar_scm.service.in > $(DESTDIR)$(mylibdir)/obs_gbp.service
 
 show-python:
 	@echo "$(PYTHON)"
--- a/README.md
+++ b/README.md
@@ -130,6 +130,13 @@ a URL, fetch the sources via the SCM from the upstream repository, and
 build a tarball from it.  You can find example `appimage.yml` files
 under the `tests/fixtures/` subdirectory.
 
+### gbp
+The `obs_gbp` service can be used to create Debian source artefacts
+(.dsc, .orig.tar.gz and if non-native .debian.tar.gz or .diff.gz) from
+Git repositories, following the very popular [git-buildpackage workflow.]
+(https://honk.sigxcpu.org/piki/projects/git-buildpackage/)
+Requires git-buildpackage to be installed.
+
 ## Archive Formats
 
 ### tar
--- a/TarSCM/__init__.py
+++ b/TarSCM/__init__.py
@@ -8,6 +8,7 @@ from TarSCM.helpers    import Helpers
 from TarSCM.cli        import Cli
 from TarSCM.archive    import Tar
 from TarSCM.archive    import ObsCpio
+from TarSCM.archive    import Gbp
 from TarSCM.exceptions import OptionsError
 
 
@@ -26,6 +27,9 @@ def run():
 
     if os.path.basename(sys.argv[0]) == "snapcraft":
         _cli.snapcraft = True
+
+    if os.path.basename(sys.argv[0]) == "obs_gbp":
+        _cli.use_obs_gbp = True
 
     task_list = Tasks(_cli)
 
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -7,6 +7,8 @@ import tarfile
 import shutil
 import glob
 import locale
+import logging
+import tempfile
 
 from TarSCM.helpers import Helpers
 
@@ -229,5 +231,133 @@ class Tar(BaseArchive):
         tar.close()
 
         self.archivefile    = tar.name
+
+        os.chdir(cwd)
+
+
+class Gbp(BaseArchive):
+
+    def create_archive(self, scm_object, **kwargs):
+        """Create Debian source artefacts using git-buildpackage.
+        """
+        args = kwargs['cli']
+        version = kwargs['version']
+
+        (workdir, topdir) = os.path.split(scm_object.clone_dir)
+
+        cwd = os.getcwd()
+        os.chdir(workdir)
+
+        if not args.revision:
+            revision = 'origin/master'
+        else:
+            revision = 'origin/' + args.revision
+
+        command = ['gbp', 'buildpackage', '--git-notify=off',
+                   '--git-force-create', '--git-cleaner="true"']
+
+        # we are not on a proper local branch due to using git-reset but we
+        # anyway use the --git-export option
+        command.extend(['--git-ignore-branch',
+                        "--git-export=%s" % revision])
+
+        # gbp can load submodules without having to run the git command, and
+        # will ignore submodules even if loaded manually unless this option is
+        # passed.
+        if args.submodules:
+            command.extend(['--git-submodules'])
+
+        # create local pristine-tar branch if present
+        ret, output = self.helpers.run_cmd(['git', 'rev-parse', '--verify',
+                                            '--quiet', 'origin/pristine-tar'],
+                                           cwd=scm_object.clone_dir)
+        if not ret:
+            ret, output = self.helpers.run_cmd(['git', 'update-ref',
+                                                'refs/heads/pristine-tar',
+                                                'origin/pristine-tar'],
+                                               cwd=scm_object.clone_dir)
+            if not ret:
+                command.append('--git-pristine-tar')
+            else:
+                command.append('--git-no-pristine-tar')
+        else:
+            command.append('--git-no-pristine-tar')
+
+        # Prevent potentially dangerous arguments from being passed to gbp,
+        # e.g. via cleaner, postexport or other hooks.
+        if args.gbp_build_args:
+            build_args = args.gbp_build_args.split(' ')
+            safe_args = re.compile(
+                '--git-verbose|--git-upstream-tree=.*|--git-no-pristine-tar')
+            p = re.compile('--git-.*|--hook-.*|--.*-hook=.*')
+
+            gbp_args = [arg for arg in build_args if safe_args.match(arg)]
+            dpkg_args = [arg for arg in build_args if not p.match(arg)]
+
+            ignored_args = list(set(build_args) - set(gbp_args + dpkg_args))
+            if ignored_args:
+                logging.info("Ignoring build_args: %s" % ignored_args)
+            command.extend(gbp_args + dpkg_args)
+
+        # Set the version in the changelog. Note that we can't simply use
+        # --source-option=-Dversion=$ver as it will not change the tarball
+        # name, which means dpkg-source -x pkg.dsc will fail as the names
+        # and version will not match
+        cl_path = os.path.join(scm_object.clone_dir, 'debian', 'changelog')
+        if (os.path.isfile(cl_path)
+                and version not in ['', '_none_', '_auto_', None]):
+            # Some characters are legal in Debian's versions but not in a git
+            # tag, so they get substituted
+            version = re.sub(r'_', r'~', version)
+            version = re.sub(r'%', r':', version)
+            with open(cl_path, 'r') as cl:
+                lines = cl.readlines()
+            old_version = re.search(r'.+ \((.+)\) .+', lines[0]).group(1)
+            # non-native packages MUST have a debian revision (-xyz)
+            if (re.search(r'-', old_version) is not None
+                    and re.search(r'-', version)) is None:
+                logging.warning("Package is non-native but requested version"
+                                " %s is native! Ignoring.", version)
+            else:
+                with open(cl_path, 'w+') as cl:
+                    # A valid debian changelog has 'package (version) release'
+                    # as the first line, if it's malformed we don't care as it
+                    # will not even build
+                    logging.debug("Setting version to %s", version)
+                    # gbp by default complains about uncommitted changes
+                    command.append("--git-ignore-new")
+                    lines[0] = re.sub(r'^(.+) \(.+\) (.+)',
+                                      r'\1 (%s) \2' % version, lines[0])
+                    cl.write("".join(lines))
+
+        logging.debug("Running in %s", scm_object.clone_dir)
+
+        self.helpers.safe_run(command, cwd=scm_object.clone_dir)
+
+        # Use dpkg to find out what source artefacts have been built and copy
+        # them back, which allows the script to be future-proof and work with
+        # all present and future package formats
+        sources = self.helpers.safe_run(['dpkg-scansources', workdir],
+                                        cwd=workdir)[1]
+
+        FILES_PATTERN = re.compile(
+            r'^Files:(.*(?:\n .*)+)', flags=re.MULTILINE)
+        for match in FILES_PATTERN.findall(sources):
+            logging.info("Files:")
+            for line in match.strip().split("\n"):
+                fname = line.strip().split(' ')[2]
+                logging.info(" %s", fname)
+                input_file = os.path.join(workdir, fname)
+                output_file = os.path.join(args.outdir, fname)
+
+                if (args.gbp_dch_release_update
+                        and fnmatch.fnmatch(fname, '*.dsc')):
+                    # This tag is used by the build-recipe-dsc to set the OBS
+                    # revision: https://github.com/openSUSE/obs-build/pull/192
+                    logging.debug("Setting OBS-DCH-RELEASE in %s", input_file)
+                    with open(input_file, "a") as dsc_file:
+                        dsc_file.write("OBS-DCH-RELEASE: 1")
+
+                shutil.copy(input_file, output_file)
 
         os.chdir(cwd)
--- a/TarSCM/scm/base.py
+++ b/TarSCM/scm/base.py
@@ -251,14 +251,18 @@ class Scm():
             'sslverify' not in self.args.__dict__ or \
             self.args.__dict__['sslverify']
 
-    def version_iso_cleanup(self, version):
+    def version_iso_cleanup(self, version, debian=False):
         """Reformat timestamp value."""
         version = re.sub(r'([0-9]{4})-([0-9]{2})-([0-9]{2}) +'
                          r'([0-9]{2})([:]([0-9]{2})([:]([0-9]{2}))?)?'
                          r'( +[-+][0-9]{3,4})',
                          r'\1\2\3T\4\6\8',
                          version)
-        version = re.sub(r'[-:]', '', version)
+        # avoid removing "-" for Debian packages, which use it to split the
+        # upstream vs downstream version
+        # for RPM it has to be stripped instead, as it's an illegal character
+        if not debian:
+            version = re.sub(r'[-:]', '', version)
         return version
 
     def prepare_working_copy(self):
--- a/TarSCM/tasks.py
+++ b/TarSCM/tasks.py
@@ -201,16 +201,23 @@ class Tasks():
 
         detected_changes = scm_object.detect_changes()
 
-        scm_object.prep_tree_for_archive(args.subdir, args.outdir,
-                                         dstname=dstname)
-        self.cleanup_dirs.append(scm_object.arch_dir)
+        if not args.use_obs_gbp:
+            scm_object.prep_tree_for_archive(args.subdir, args.outdir,
+                                             dstname=dstname)
+            self.cleanup_dirs.append(scm_object.arch_dir)
 
+        # For the GBP service there is no copy in arch_dir, so use clone_dir
+        # which has the same content
+        extract_src = scm_object.arch_dir
         if args.use_obs_scm:
             arch = TarSCM.archive.ObsCpio()
+        elif args.use_obs_gbp:
+            arch = TarSCM.archive.Gbp()
+            extract_src = scm_object.clone_dir
         else:
             arch = TarSCM.archive.Tar()
 
-        arch.extract_from_archive(scm_object.arch_dir, args.extract,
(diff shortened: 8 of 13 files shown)


adrianschroeter:
hm, I am not directly against merging, but we won't deploy this on our server most likely, since we want to go away from tar balls anyway.

Can't you just make the execution of git-buildpackage an option of "obs_scm" service? Can it operate on a checked out git tree?

bluca:
It was an option at the beginning, but then I realised that the tar/cpio were created and thrown away immediately, so it was a waste of time and I/O.

I work a lot with the kernel, and each unneeded step adds minutes to the service run time since it's a massive repository :-/ Same for Qemu and other large projects.

I understand the goal of moving away from tarballs.

The problem with the current scm service and debian packages, as I've written in the commit message, is that it needs a lot of massaging and ad-hoc changes specifically and only for OBS and really only works because it's forcing a downgrade in the package format to the deprecated 1.0. That is eventually going away, and when that happens I'm afraid it won't work anymore - a non-native (quilt) 3.0 package needs a .dsc, an orig tarball and a debian tarball.

This service instead allows to build debian packages straight from the debian git repositories, without any OBS-specific changes, which I think is nice :-)

TarSCM/scm/git.py
```diff
@@ -251,7 +251,7 @@ def prepare_working_copy(self):
         # between multiple services
         org_clone_dir  = self.clone_dir
         self.clone_dir = self.repodir
-        command = ['git', 'clone', self.repocachedir, self.url, self.clone_dir]
+        command = ['git', 'clone', '--no-checkout']
```
M0ses: @adrianschroeter  I`m convinced that this would break the workflow for locally cached changes.

M0ses:
> It was an option at the beginning, but then I realised that the tar/cpio were created and thrown away immediately, so it was a waste of time and I/O.

I do not understand what you mean, could you explain the used services and workflow a bit more detailed?

TarSCM/scm/git.py
bluca: What workflow is that exactly? Isn't this clone happening in an ephemeral tmpdir? If there's another way this is used please let me know and I'm happy to fix or remove

bluca:
> I do not understand what you mean, could you explain the used services and workflow a bit more detailed?

Certainly - the most common git workflow used by Debian developers makes use of git buildpackage (full documentation: https://honk.sigxcpu.org/piki/projects/git-buildpackage/ ).
In the case of non-native packages, which is where it's most useful, the upstream code is stored either unpackacked in an "upstream" branch, or as pristine-tar deltas in a "pristine-tar" branch.
At source build time, git-buildpackage puts these back together (in various ways and with various options) to create the 3 source artefacts needed by dpkg-buildpackage to do the actual binary build: the dsc file, the orig.tar.gz archive and the debian.tar.gz archive. They are all necessary for the 3.0 format, which is the current packaging format.
Especially note that the dsc file is _never_ checked in git - it's the output of a build, not the input.

The cpio/tar archives created by obs/tar_scm are not useful by themselves as they lack the above artefacts. And they are also redundant, as gbp creates all 3 artefacts, hence why I went with a third option - so that the redundant unpack + repack steps, which in the case of large repositories takes minutes and loads of I/O, could be skipped. Also I found it much clearer this way - it is for all intent and purposes a third archiving option, specific to make the Debian workflow much easier and without OBS-specific quirks.

TarSCM/scm/git.py
M0ses: If you do a "osc build" on your local checkout, the git repo is cached. You can do your changes in this local cache and rerun "obs build". The local changes are used for these builds. When you are done, you simply commit those changes to your git repo and do a "osc service rr" and the obs server takes your changes from your git repo, builds a package and publishes the new package.

This would be broken if we would do '--no-checkout' in general.

M0ses:
Hmm, obscpio's are intended to reduce the disk usage in OBS. This might not be a problem for a private instance, but for a "cloud service" like build.opensuse.org with about 50000 Users it is.

So first of all, I would like to understand if 
* you only want to get the new format included in obs-service-tar_scm
* you would like to get it deployed on build.opensuse.org
* it makes sense to discuss other solutions that fits both needs (private instance vs. build.opensuse.org)

TarSCM/scm/git.py
bluca: Thanks, I'll try that now.

As far as I can see this function is only called here:
https://github.com/openSUSE/obs-service-tar_scm/blob/master/TarSCM/scm/base.py#L80
And immediately afterwards a git reset --hard is done to the selected revision (or the program errors out if the revision is not found). So it looked to me like the checkout was redundant and would be immediately overwritte, and in theory it shouldn't affect that workflow either. But I'll check to make sure.

bluca: So I did:

```
osc co home:bluca:branches:network:messaging:zeromq:git-draft/zproject
cd home\:bluca\:branches\:network\:messaging\:zeromq\:git-draft/zproject
osc service localrun
osc build
echo lol >> zproject/zproject_android.gsl
osc build
```

And the "lol" is there in the built package at the bottom of the file.
The _service uses tar_scm: https://build.opensuse.org/package/show/home:bluca:branches:network:messaging:zeromq:git-draft/zproject

Did I miss something?

bluca:
I understand the need to reduce resource usage, and absolutely agree it's very important for a public service, no questions there.

I do have a private instance at work, where I would mostly take advantage of this, so having the new format included even if it's not deployed would be already very nice!

But in general, as a Debian maintainer, I find the available workflows to build Debian packaging a bit clunky as I wrote in the other comments. It's perfectly understandable that SUSE optimises for RPM given it's what it uses, but I do believe there's scope for improvements, and I think it would be very nice for other users of the public instance to have them available as well :-)

So if I understand the idea behind the cpio, it's that within the same package it stores an incremental diff in the revision history rather than the full copy like a tarbal, is that right?

What if tarballs too were stored as deltas, for example using pristine-tar? Have you already explored this solution when the cpio work was done?

TarSCM/scm/git.py
M0ses: No, AFAICS it works as expected :-). I think the "osc localrun" might be not necessary but anyway this is how I would have expected. We should check if there is already a test case for this and document it here.

M0ses:
> I understand the need to reduce resource usage, and absolutely agree it's very important for a public service, no questions there.
> 
> I do have a private instance at work, where I would mostly take advantage of this, so having the new format included even if it's not deployed would be already very nice!
> 

OK, lets keep this in mind as a "fallback" :-)

> But in general, as a Debian maintainer, I find the available workflows to build Debian packaging a bit clunky as I wrote in the other comments. 

I think you meant working with debian package in OBS, right?

>It's perfectly understandable that SUSE optimises for RPM given it's what it uses, but I do believe there's scope for improvements, and I think it would be very nice for other users of the public instance to have them available as well :-)

OBS always had the goal to be a universal solution for building package for linux. Indepdent from package formats, distributions, etc.. So we are always open to reasonable solutions which fit into our agenda and are happy to get valuable contributions :-)

> So if I understand the idea behind the cpio, it's that within the same package it stores an incremental diff in the revision history rather than the full copy like a tarbal, is that right?

Yes, the idea is to store deltas independently from the used VCS

> What if tarballs too were stored as deltas, for example using pristine-tar? Have you already explored this solution when the cpio work was done?

I was not aware of pristine-tar so far, but maybe @mlschroe, who developed bs_deltastore, could shed light on me.

bluca:
Yes, I meant working with Debian packages in OBS :-)
With RPM is nice and easy, just have a spec and it's done. It can work from a common git workflow out of the box. With Debian packages there's a lot of hoops to jump through - have a fake dsc file to feed the scheduler/resolver, extract the changelog and rename it to get debtransform to change the revision, force-overwrite the packaging format to the deprecated 1.0. It cannot work from a common git workflow out of the box.

So in terms of moving forward, what do you think of:

1) If there are no objections/comments, merge this so myself and other people can deploy in their own private instances if they wish
2) Do not deploy on the SUSE public instance
3) I'll try and figure out if the tarballs can be stored as deltas, like with cpio. If I am successfull I'll propose that change and if you guys like it and meets your requirements we can discuss deployment again, if you are interested.

How does this sound?

aspiers:
I'm keen to help here (even though I work for SUSE, I like and use Debian too ;-) ), but sorry - as I'm relatively ignorant about the Debian packaging process I really need help understanding this PR.  Please can we step back for a moment?  You wrote:

> Debian + git developers workflow never include the .dsc file in the repository, as it is the output of a build, not the input. This makes it awkward and convoluted to use tar_scm with Debian git repositories, and just unfeasable with obs_scm.

I don't see why the first sentence implies the second sentence.  So I looked for clarification in your later comments, and found this:

> At source build time, git-buildpackage puts these back together (in various ways and with various options) to create the 3 source artefacts needed by dpkg-buildpackage to do the actual binary build: the dsc file, the orig.tar.gz archive and the debian.tar.gz archive. They are all necessary for the 3.0 format, which is the current packaging format. Especially note that the dsc file is never checked in git - it's the output of a build, not the input.

I must be misunderstanding something because to me this paragraph seems to contradict itself.  The first sentence says that the `.dsc` is one of the 3 source artefacts required to build the binary packages, i.e. it's an input to the build. But then the third sentence says that it's the output of a build, not the input.  So which is it, an input or output?  I don't see how it could be both, because then you'd have a catch 22 situation, unless somehow you bootstrapped with a dummy `.dsc` file, did a build, and then built again with the resulting new `.dsc` ... but that sounds crazy, so I doubt it's what happens.  And according to the [Debian New Maintainers' Guide](https://www.debian.org/doc/manuals/maint-guide/build.en.html) it's an output not an input, which is generated from the `control` file.  So while I see the need for the `control` file to be available at build-time, I don't see the need for the `.dsc` file at build-time.  Please help eliminate my confusion! :-)

> > What if tarballs too were stored as deltas, for example using pristine-tar? Have you already explored this solution when the cpio work was done?
> I was not aware of pristine-tar so far, but maybe @mlschroe, who developed bs_deltastore, could shed light on me.

FTR [`pristine-tar` was mentioned briefly on an internal SUSE list a long time ago](https://mailman.suse.de/mlarch/SuSE/research/2015/research.2015.10/msg00064.html) (apologies for the SUSE-only link).

bluca:
Thanks for following up - I appreciate that.

> I must be misunderstanding something because to me this paragraph seems to contradict itself. The first sentence says that the .dsc is one of the 3 source artefacts required to build the binary packages, i.e. it's an input to the build. But then the third sentence says that it's the output of a build, not the input. So which is it, an input or output? I don't see how it could be both, because then you'd have a catch 22 situation, unless somehow you bootstrapped with a dummy .dsc file, did a build, and then built again with the resulting new .dsc ... but that sounds crazy, so I doubt it's what happens. And according to the Debian New Maintainers' Guide it's an output not an input, which is generated from the control file. So while I see the need for the control file to be available at build-time, I don't see the need for the .dsc file at build-time. Please help eliminate my confusion! :-)

Sorry for the confusion! Let me try to clarify a bit.

As you correctly pointed out the dsc + orig.tar.gz (+ debian.tar.gz for non-native) source artefacts are the output of a source build, and should be the equivalent of the src.rpm, if that helps.

What I meant to say is that those source artefacts are then the input of the **OBS build** process - if you try to upload just the repository content, without a dsc, OBS will refuse to build it. The dsc file is required by build-recipe-dsc in obs-build.

So it's chained as: repository --> dsc + orig.tar.gz (+ debian.tar.gz)  --> obs-build

Note that there is a similar process in the Debian toolchain - where the source artefacts are fed to the buildd network.

So if you just have a git repository, and create a tarball/cpio from it, that's not enough to do anything in OBS for Debian packages.
It works for RPM, because the spec gets extracted and that's it, but not for Debian as there's no dsc to extract in a normal git repository.
But for Debian it doesn't - and the git-buildpackage tool is commonly used to bridge that gap, and get the source artefacts from a git repository. There's a similar one for SVN as well (and some weird stuff for BZR :-) )

The most common workaround I've seen in OBS is to create a "fake" dsc file, hardcoding the expected values. But it's completely non-standard so one has to use an OBS-specific git repository, and it needs to be maintaned manually as build dependencies change, so it's way less than optimal.

aspiers:
Sorry, I only just saw this reply.  Not getting the right notifications from GitHub, no idea why :angry:

bluca:
@aspiers no worries! I'm also on IRC in #opensuse-buildservice during GMT working hours if you have more questions and want to chat live.

aspiers:
Including relevant IRC conversation here for the record:

```
2018-01-02 12:04:26 <aspiers>	bluca: are you around?
2018-01-02 12:04:36 <bluca>	aspiers: hi
2018-01-02 12:04:41 <aspiers>	hi
2018-01-02 12:04:49 <aspiers>	firstly sorry for the delay in reviews
2018-01-02 12:05:01 <bluca>	no worries :)
2018-01-02 12:05:05 <aspiers>	I'm trying to finish them now
2018-01-02 12:05:29 <bluca>	great, let me know if you'd like any other changes
2018-01-02 12:05:49 <aspiers>	part of the problem is that despite your much-appreciated efforts to explain it, I am still confused by the Debian packaging process ;-/
2018-01-02 12:06:01 <aspiers>	e.g. https://github.com/openSUSE/obs-service-tar_scm/pull/179#issuecomment-319724490
2018-01-02 12:06:04 <bluca>	eheh
2018-01-02 12:06:11 <bluca>	you are not the only one ;-)
2018-01-02 12:06:25 <bluca>	note that I split the bug fixes in another PR: https://github.com/openSUSE/obs-service-tar_scm/pull/209
2018-01-02 12:06:29 <aspiers>	you seem to be saying that .dsc files are both outputs from a build and inputs to a build, which I cannot understandd
2018-01-02 12:06:45 <aspiers>	yes I saw that, but I need to understand the motivation for the changes, so I went to the original PR
2018-01-02 12:07:10 <aspiers>	I'm trying to understand, where do .dsc files come from? are they manually written, or auto-generatedd?
2018-01-02 12:07:30 <bluca>	generated
2018-01-02 12:07:45 <bluca>	so in the git workflow, you have the source in the repository
2018-01-02 12:08:02 <bluca>	then you do a source build (eg: dpkg-buildpackage -S)
2018-01-02 12:08:12 <bluca>	that gives you orig.tar.gz + debian.tar.gz + .dsc
2018-01-02 12:08:38 <bluca>	the dsc is generated, never checked in
2018-01-02 12:08:42 <bluca>	and that's the crux of the problem
2018-01-02 12:09:00 <aspiers>	who does a source build though?
2018-01-02 12:09:07 <bluca>	then typically that triad is uploaded to the FTP, and then picked up by various builders to produce the actual binary packages
2018-01-02 12:09:23 <bluca>	developers
2018-01-02 12:10:51 <aspiers>	OK so it's fundamentally a different process to rpm, where developers don't need to do any type of build before uploading to OBS
2018-01-02 12:11:17 <bluca>	yes, because all OBS needs for an RPM binary builds is the spec file plus the sources
2018-01-02 12:11:22 <bluca>	and the spec file is typically checked in git
2018-01-02 12:11:32 <aspiers>	actually it's not that typical
2018-01-02 12:11:44 <bluca>	ok, sometimes it is :)
2018-01-02 12:11:49 <aspiers>	it's not uncommon :)
2018-01-02 12:12:00 <aspiers>	but also very common to have the .spec only in OBS
2018-01-02 12:12:38 <bluca>	I've seen many cases where developers on OBS just have an obs_scm service, and everything in git
2018-01-02 12:12:52 <aspiers>	sure
2018-01-02 12:13:00 <bluca>	that works nicely for RPMs, but for DEBs there's this problem
2018-01-02 12:13:11 <bluca>	ultimately I think the issue is that obs-build needs a dsc file
2018-01-02 12:13:24 <aspiers>	so every piece of Debian software requires *two* builds: a manual build, then a subsequent automatic build?
2018-01-02 12:13:28 <bluca>	while it should be able to look for the information it needs (build depndencies mostly I think) from debian/control too
2018-01-02 12:13:31 <aspiers>	that seems weird
2018-01-02 12:13:56 <aspiers>	yes, from my limited knowledge debian/control is similar to .spec, right?
2018-01-02 12:14:00 <aspiers>	i.e. manually written metadata
2018-01-02 12:14:16 <aspiers>	so it seems wrong to me that OBS can't build from that
2018-01-02 12:14:18 <bluca>	yeah d/control has the manually written metadata that typically is at he beginning of as pec
2018-01-02 12:14:55 <bluca>	it seems weird because it can get even more complicated :-P
2018-01-02 12:15:25 <aspiers>	so I was looking at the first date-formatting commit
2018-01-02 12:15:46 <jimklimov>	and in bluca's case, the control files are (can be) generated too - but are SCM-tracked ;)
2018-01-02 12:15:54 <bluca>	yes the problem there is that it mistakenly identifies a debian version, eg: 3.0.1-2 as a date
2018-01-02 12:15:59 <bluca>	and it removes the "-"
2018-01-02 12:16:01 <aspiers>	what worries me is that your change decides what to do without any consideration for whether it's a Debian nor non-Debian context
2018-01-02 12:16:02 <bluca>	which breaks many things
2018-01-02 12:16:24 <bluca>	as far as I understand from that bit, it's trying to format dates
2018-01-02 12:16:33 <bluca>	so I though - if it's not a date, don't do anything
2018-01-02 12:16:55 <bluca>	does that sound reasonable as an approach?
2018-01-02 12:16:57 <aspiers>	ah, think I misread the diff
2018-01-02 12:17:01 <aspiers>	thinking...
2018-01-02 12:20:13 <bluca>	basically, the issue with that code is the second re.sub
2018-01-02 12:20:23 <aspiers>	yeah
2018-01-02 12:20:52 <aspiers>	but that second re.sub is absolutely needed for rpms, where versions are not allowed to contain '-'
2018-01-02 12:20:52 <bluca>	as far as I could understand from the code, its purpose was again just to work on dates
2018-01-02 12:21:12 <aspiers>	so I'm worried that removing it could break some packages
2018-01-02 12:21:21 <bluca>	mh I didn't know that, thanks for pointing it out
2018-01-02 12:21:31 <bluca>	the comment and code don't make it obvious :P
2018-01-02 12:21:43 <aspiers>	there are plenty of things about this code which are not obvious :-/
2018-01-02 12:21:46 <bluca>	"""Reformat timestamp value."""
2018-01-02 12:21:53 <aspiers>	I know
2018-01-02 12:22:00 <aspiers>	I didn't write it ;-)
2018-01-02 12:22:12 <bluca>	yeah iirc it was that jblunck guy, always breaking stuff he is :P
2018-01-02 12:22:38 <bluca>	(we were colleagues at vyatta for 2 years)
2018-01-02 12:22:39 <aspiers>	lol
2018-01-02 12:23:02 <aspiers>	blame(someone_else);
2018-01-02 12:23:08 <bluca>	eheh
2018-01-02 12:23:27 <bluca>	so how do we get around this? the issue is that iirc there is no knowledge of what the code is actually acting on at that point
2018-01-02 12:23:36 <aspiers>	good question
2018-01-02 12:23:40 <bluca>	or in most parts of obs_scm iirc
2018-01-02 12:24:04 <bluca>	this is quite a fundamental "disagreement" - leaving a - breaks rpm, but removing it breaks debs
2018-01-02 12:24:17 <aspiers>	right ;-/
2018-01-02 12:24:42 <aspiers>	so what's the schema for Debian packages? rpm is NAME-VERSION-RELEASE
2018-01-02 12:24:47 <aspiers>	I thought Debian was similar
2018-01-02 12:25:19 <aspiers>	VERSION corresponds to upstream version, RELEASE is downstream build (taking into account patches or other changes to .spec, rebuilds etc.)
2018-01-02 12:26:08 <bluca>	NAME[-MORE]_UPSTREAM_VER-DEB_VER_ARCH
2018-01-02 12:26:15 <aspiers>	IIRC in OBS, RELEASE is X.Y where X is incremented when sources change and Y is reset, or Y is incremented on rebuilds
2018-01-02 12:26:23 <jimklimov>	http://manpages.ubuntu.com/manpages/zesty/man5/deb-version.5.html
2018-01-02 12:26:33 <bluca>	actually more complicated than that, but yea
2018-01-02 12:26:41 <bluca>	yep thanks jim
2018-01-02 12:26:54 <aspiers>	ahhh of course, I forgot it uses _ to delimit :-/
2018-01-02 12:27:37 <aspiers>	I have Debian and Ubuntu machines for my personal use, but I don't do much with packages on them except apply updates :)
2018-01-02 12:27:39 <bluca>	actually OBS does Y.X which is mightly annoying :-) as same version re-checked out sorts out as a older than the previous checkout that gets rebuilt
2018-01-02 12:27:44 <bluca>	but I digress :P
2018-01-02 12:27:55 <aspiers>	really?!
2018-01-02 12:28:10 <jimklimov>	in our project (that uses/collaborates with bluca's technology) we sort of abuse this to produce versions like 1.0.0.20171229024334~git6eb8d07cdc9f-1
2018-01-02 12:28:21 <jimklimov>	which is valud as far as apt/dpkg are concerned
2018-01-02 12:28:34 <aspiers>	yes, similar (ab)use is common with rpm
2018-01-02 12:28:37 <bluca>	I'm pretty sure yeah - at least obs-build-dsc does that
2018-01-02 12:28:52 <bluca>	(off work so can't actually log in our instance to double check)
2018-01-02 12:29:00 <jimklimov>	this allows us to track (rarely-changing) nominal versions like 1.0.0 matched to the actual git checkout timestamp~hash
2018-01-02 12:29:18 <jimklimov>	and the respin number I think is hardcoded to "-1" here
2018-01-02 12:29:25 <bluca>	anyway yeah it seems to me the fundamental issue is that in the rpm world the upstream and downstream versions are separated, and obs handles them as uch
2018-01-02 12:29:32 <aspiers>	e.g. https://build.opensuse.org/package/binaries/Cloud:OpenStack:Master/openstack-nova-doc?repository=SLE_12_SP3
2018-01-02 12:29:35 <bluca>	while in the deb world there is one version field
2018-01-02 12:30:00 <aspiers>	but there is a hyphen delimiter within that field?
2018-01-02 12:30:25 <aspiers>	which the deb tools ignores?
2018-01-02 12:31:47 <bluca>	yes the hyphen splits the upstream vs downstream
2018-01-02 12:32:11 <aspiers>	checking I understand: is it ${NAME}[-${MORE}]_${UPSTREAM_VER}-${DEB_VER}_ARCH ?
2018-01-02 12:32:12 <bluca>	but the vast majority of tools do not care about that split
2018-01-02 12:32:27 <aspiers>	ok
2018-01-02 12:32:30 <bluca>	not sure how the rpm ones work, but apt will treat the whole thing as the version number
2018-01-02 12:32:35 <bluca>	for example
2018-01-02 12:32:58 <aspiers>	It's definitely separate in rpm. The .spec has separate Version: and Release headers
2018-01-02 12:33:19 <bluca>	yes that's the scheme for .deb package names
2018-01-02 12:33:35 <bluca>	right, remember seeing those
2018-01-02 12:34:21 <aspiers>	OK
2018-01-02 12:34:34 <aspiers>	so your other 2 commits in that PR are unrelated and totally non-controversial
2018-01-02 12:34:43 <aspiers>	I'd be happy to merge those immediately
2018-01-02 12:35:09 <aspiers>	can you split them (or the date thing, it doesn't matter which) off into a separate PR?
2018-01-02 12:35:42 <bluca>	ok, thanks - just force pushed to that PR
2018-01-02 12:35:44 <aspiers>	if I had published my git-explode utility by now, you could have done that as a one-liner :)
2018-01-02 12:35:55 <aspiers>	although I did already publish git-transplant IIRC
2018-01-02 12:42:13 <aspiers>	bluca: regarding the date/version issue, I'd recommend talking to AdrianS and m0ses when they get back - they will probably have some good suggestions if you can't think of any solution yourself
2018-01-02 12:42:45 <bluca>	yes I'm re-looking at the code right now, I'll re-assess with this new info
2018-01-02 12:42:51 <bluca>	thanks for that, as I didn't know of this problem
2018-01-02 12:43:02 <aspiers>	bluca: actually I'm still confused ... the .dsc file should be checked into the OBS project right?
2018-01-02 12:43:13 <aspiers>	so can't the version come from there?
2018-01-02 12:44:42 <aspiers>	I've got to nip out to drop a parcel off, back in a bit
2018-01-02 12:47:49 <bluca>	I think at that point the obs_scm service code has just access to what's in git
2018-01-02 12:48:33 <bluca>	also if the end goal is to be able to support a pure git workflow - everything in git, with the sources generated by that service - then it might not even be available
2018-01-02 13:07:34 <bluca>	so I've re-set up the example and the repro
2018-01-02 13:08:14 <bluca>	the formatting problem is mainly an issue with the new obs_gbp from the other pr
2018-01-02 13:08:34 <bluca>	commonly the gbp workflow tags versions as eg: debian/3.0.1-1
2018-01-02 13:09:06 <bluca>	in the example I've written in the PR, it can be seen that without the formatting fix obs_scm will mangle that to 3.0.11
2018-01-02 13:09:36 <bluca>	the good news is that outside of the gbp workflow, I don't think this problem will be too much important
2018-01-02 13:09:45 <bluca>	so the 2 changes can be tied together
2018-01-02 13:10:51 <bluca>	brb lunchtime :-)
2018-01-02 14:13:11 <bluca>	aspiers: to be precise, users of obs_scm + debian are unlikely to have a problem given they already need to apply OBS-specific workarounds to their git repo
2018-01-02 14:13:29 <bluca>	like myself and jimklimov already do
2018-01-02 14:13:58 <bluca>	so not much point in spending time on it, unless you are willing to consider the obs_gbp enhancement :-)
```

nickbroon:
Just spotted this. This would be great to have! Would save having to manually create the .dsc and tarballs and adding them to OBS.

bluca:
@aspiers I have pushed a new version, which solves the version stripping problem we talked about - (RPM needs to remove - and : but Debian needs them): now the removal of - and : is skipped if it's the new obs_gbp service that is being ran. This way RPM repositories will not be affected, and Debian packages won't have a broken version.
Does it sound good as a solution?

bluca:
@adrianschroeter @aspiers ping :-) Any chance we could bring forward this discussion and un-stall it? Thanks!

nickbroon:
@adrianschroeter @aspiers What is the state of this Pull Request? Can the obs_gbp service be added? It really will make the Debian package flow much easier.

adrianschroeter:
Just as an update, I am still fine in merging it, but we won't offer it on the service server of build.opensuse.org.

The reason behind is that we want to go away from tar balls, where we can not store the delta.

So we can merge it now, but it would be cool if it could be splitted into first storeing .obscpio archives and create the tar balls at buildtime with another service.

bluca:
> Just as an update, I am still fine in merging it, but we won't offer it on the service server of build.opensuse.org.
>
> The reason behind is that we want to go away from tar balls, where we can not store the delta.
>
> So we can merge it now, but it would be cool if it could be splitted into first storeing .obscpio archives and create the tar balls at buildtime with another service.

Thanks Adrian, sounds good to me. Pushed to resolve merge conflicts and Hound warnings.

I have since moved on, so I'll ask former colleagues who I believe are still using this if they are up for the follow up work to integrate with obscpio.

TarSCM/cli.py
```diff
@@ -191,6 +199,9 @@ def verify_args(self, args):
         args.package_meta    = bool(args.package_meta == 'yes')
         args.sslverify       = bool(args.sslverify != 'disable')
         args.use_obs_scm     = bool(args.use_obs_scm)
+        args.use_obs_gbp     = bool(args.use_obs_gbp)
+        args.gbp_dch_release_update = bool(args.gbp_dch_release_update
+                                           != 'disable')
```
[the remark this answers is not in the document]
bluca: The bot doesn't like line breaks before or after binary ops... what's the preferred format @adrianschroeter ?

nickbroon:
I'd be happy to follow up this PR with investigating if gbp/dpkg-source can generate .cpio instead of .tar.gz .tar.xz, of if the new obs_gbp service can covert these into .cpio files. The complexity is that .dsc file generated by the source service references the tarballs and has checksums for them, so later regenerating these tarballs from the cpio files might be complex to get correct. Ultimately this might complex enough that it would required gbp/dpkg-source to natively support cpio themselves.... 
But for the moment it would be great to get this merged as we could make use of it on our OBS server.

bluca:
> I'd be happy to follow up this PR with investigating if gbp/dpkg-source can generate .cpio instead of .tar.gz .tar.xz, of if the new obs_gbp service can covert these into .cpio files. The complexity is that .dsc file generated by the source service references the tarballs and has checksums for them, so later regenerating these tarballs from the cpio files might be complex to get correct. Ultimately this might complex enough that it would required gbp/dpkg-source to natively support cpio themselves....
> But for the moment it would be great to get this merged as we could make use of it on our OBS server.

I'm a bit rusty, but IIRC the checksums don't matter at service runtime, so I think this is what Adrian is proposing and it should work in theory:

service runs -> obs_gbp creates tarball + dsc -> convert tarball to obscpio -> dsc is looked at by scheduler -> dsc + obscpio given to builder -> build time service converts obscpio back to tarball before dpkg-build runs

Both formats are reproducible, so with the right flags it should be possible to get back the exact same tarball, I think.

nickbroon:
> service runs -> obs_gbp creates tarball + dsc -> convert tarball to obscpio -> dsc is looked at by scheduler -> dsc + obscpio given to builder -> build time service converts obscpio back to tarball before dpkg-build runs
> 
> Both formats are reproducible, so with the right flags it should be possible to get back the exact same tarball, I think

That's what I also had in mind, I was perhaps just more nervous on the conversion of the obscpio back into a tarball in a reproducible way. But I'll test that this can be done, and if the checksums can be ignored then all the better. Thanks @bluca

PR #182, opened 2017-07-26
also override timestamps of files in obscpio

bmwiedemann:
to allow a more similar cpio to be produced
from git checkouts on different hosts

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -88,7 +88,9 @@ class ObsCpio(BaseArchive):
                 if not METADATA_PATTERN.match(name):
                     cpiolist.append(name)
 
+        tstamp = self.helpers.get_timestamp(scm_object, args, topdir)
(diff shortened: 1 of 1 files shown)


aspiers:
@M0ses If this looks good to you, please merge. Thanks!

bmwiedemann:
rebased it to make the diff easier to read now that #181 is merged.

aspiers:
Thanks @bmwiedemann, still +1 from my side.

aspiers:
I guess @M0ses is on vacation? @bmwiedemann If this is blocking you, we can just merge without waiting since it seems low-risk to me.

bmwiedemann:
It has some risk, because we are changing timestamps in the git checkout, which might matter if people use make in there. Also, this patch is not that important, because we usually only produce .obscpio files once and submit them as blobs through OBS

PR #187, opened 2017-08-29
change ordering so that latest change is on top

bmwiedemann:
This makes diffing of rpm changelogs a lot easier
when different packages pull changes from related git repos

E.g. when upstream does commits
A B C D E
and one package pulls after B and E
and another package pulls after C and E,
both should have at the end a changes file with
E
D
C
B
A

but without this patch, the first one has
C
D
E
A
B

and the 2nd one
D
E
A
B
C

reverse was maybe introduced in commit 88e9d57a by @jblunck

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -234,7 +234,7 @@ class Git(Scm):
 
         logging.debug(dbg_msg)
 
-        lines = self._log_cmd(['--reverse', '--no-merges',
+        lines = self._log_cmd(['--no-merges',
                                '--pretty=format:%s',
                                "%s..%s" % (last_rev, current_rev)],
                               subdir)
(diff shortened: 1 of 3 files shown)


jblunck:
This was added by @aspiers on purpose:

https://github.com/openSUSE/obs-service-tar_scm/commit/9acba1526201a60c026f25fd8666db6e3deb95a6

You should reference the correct commit in the log message.

aspiers:
Yeah, thanks for digging out the history @jblunck - when I wrote 9acba15 I guess I was probably only concerned with preserving the behaviour of the old shell script.  However as pointed out in this PR, the old behaviour was illogical, so we should (re-)revert to the newest-first format originally implemented by @jblunck and proposed again here.

bmwiedemann:
I thought, 9acba15 just moved the 'reverse' from line 642 to 639

adapted the fixtures

aspiers:
@bmwiedemann I think tests are now failing because they are for both git and svn, and the svn code is still generating changes with the other order - you will need to switch the svn order too. Sorry ;-)

PR #191, opened 2017-09-20
Add include_topdir option for the tar service

### LINKED ISSUE #190: Let user choose to include or not the topdir in tar archives
## Description

In KubernetesCore team we are building our Docker images using KIWI and OBS and we would like to have the sources of our KIWI project at github but we face some difficulties regarding that. Typically any KIWI project is structured as:

- _config.xml_ (the description file)
- _condif.sh_ (the configuration script)
- _root.tar.gz_ (the overlay files)

As it happens [here](https://github.com/openSUSE/docker-containers) the _root.tar.gz_ tarball contains multiple files including scripts, specially in Docker containers where it is not strange to have some entrypoint script there. We would like to have this _root.tar.gz_ file uncompressed at github as it may include some code specific for the repo which is not present in anywhere else, thus it is so convenient to be able to make PR and reviews in a regular way in that case. I have been trying to use _obs_scm_ service to build _root.tar.gz_ file on the fly, assuming a _root_ folder in the github repo with the following configuration:

```xml
<services>
    <service name="obs_scm" mode="disabled">
        <param name="url">https://github.com/davidcassany/docker-containers.git</param>
        <param name="scm">git</param>
        <param name="subdir">openSUSE-Tumbleweed</param>
        <param name="revision">switch_to_kiwi-ng-uncompressed-root</param>
        <param name="extract">config.kiwi</param>
        <param name="extract">config.sh</param>
    </service>
    <service name="tar" mode="disabled">
        <param name="subdir">openSUSE-Tumbleweed/root</param>
        <param name="filename">root</param>
        <param name="extension">tar.gz</param>
    </service>
</services>
```
This configuration creates the file _root.tar.gz_ and it is almost what we need, but there is an issue. The resulting root.tar.gz file includes the top directory (in this case the _root_ folder) in it. KIWI expands _root.tar.gz_ in `/` of the root tree image, thus in this case the top directory should not be included.

I am wondering if it would make sense to include a new parameter for the tar service in order to include or not the top directory, something like:

```xml
...
<service name="tar" mode="disabled">
    <param name="subdir">openSUSE-Tumbleweed/root</param>
    <param name="filename">root</param>
    <param name="extension">tar.gz</param>
    <param name="include_topdir">no</param>
</service>
```

Does it make any sense to you? I would expect _include_topdir_ enabled by default. I could try to provide a patch for it, but I'd like to know your opinion before investing some time on it.

davidcassany:
This commit adds include_topdir client option for the tar service.
Concretely it enables the user to choose if the top directory of the
content that has been required to tar should or not be included in the
resulting tarball. This is something revelant in order to allow OBS
services to fetch KIWI projects from some SCM.

Fixes #190


davidcassany:
@aspiers, @M0ses, did you have time to have a look at it? Not sure if you are awaiting me to take some action, I am willing to collaborate if so.

M0ses:
@davidcassany Thanks for your contribution

@davidcassany @aspiers 

Sorry for being the party pooper, but what if I want my files under x/y/z.
Shouldn`t this be supported also?

davidcassany:
> Sorry for being the party pooper, but what if I want my files under x/y/z.
Shouldn`t this be supported also?

Yes, in fact that would be nice. Do you have any suggestion of how we could express that in the _service xml file? I mean how the parameter syntax should look like. Maybe something like:

```xml
<param name="rename_topdir">x/y/z</param>
```
And then assume that the current `topdir` should be replaced by `x/y/z`? and use `_none_` in order to not include anything?

aspiers:
@davidcassany @M0ses I'm beginning to think it would be cleaner both from a UI and an implementation perspective if the parameter(s) allowed an arbitrary search-and-replace which happened within `tar_filter`.  So your use case would be covered by

    <param name="path_filter_search">^root/</param>
    <param name="path_filter_replace"></param>

and the one suggested by @M0ses would look like:

    <param name="path_filter_search">^</param>
    <param name="path_filter_replace">x/y/z/</param>

but it would also allow other potentially useful stuff like trimming a version suffix from the topdir path segment, e.g.:

    <param name="path_filter_search">^([^/]+?)-[0-9][^/]+/</param>
    <param name="path_filter_replace">\1/</param>

or relocating stuff:

    <param name="path_filter_search">^usr/</param>
    <param name="path_filter_replace">usr/local/</param>

This would only require adding a few extra lines to `tar_filter`'s definition.  What do you think?

davidcassany:
@aspiers, yes, that makes sense to me, we have just be careful not to alter the exclude feature that is currently implemented with a tar filter, but this should be easy to verify with unit tests. If I try to go this way would you prefer to have it in a separate PR or continue here in this PR?

davidcassany:
@aspiers, just found a little issue with your suggestion, tarfile python module does not support `filter` in python 2.6. I believe this should not prevent us of implementing such a feature, IMHO in case someone ever uses this service with python 2.6 it could be acceptable that related parameters are just ignored and no filter is applied, making the filtering feature only available from python 2.7. What you think?

adrianschroeter:
Just for reference, we discuss here a different solution for the same problem:

 https://github.com/openSUSE/docker-containers/pull/73

davidcassany:
As @adrianschroeter mentioned, he proposed another approach for the #190 issue which is actually working and that I am already happy with. Thus I will close #190 issue since it can be solved by a different approach, basically without trying to tar the folder from the repository (I was not aware it was possible to use a folder inside the build service this way). So, from our current needs this story is solved, the solution was already there.

That being said, we should consider if the current change/update of this service according to the last @aspiers suggestion is something that we want to implement or not. I leave the decision to the maintainers, if you are willing to include something like that you can count on me for a contribution, if you consider it is not worth is also fine, since @adrianschroeter already found a solution for my former issue.

Thanks to all

davidcassany:
@aspiers @M0ses, hi, I finally found some time to finalize this PR according to the Adam suggestions.

I found myself in situation where would be handy to relocate a file thus I believe that this is still an interesting feature for the service.

M0ses:
@davidcassany I'm missing the new cli options in the tar_scm.service.in file - how should they be configured in OBS service files?

PR #196, opened 2017-10-18
add a lot more detail to README

### LINKED ISSUE #156: snapcraft is undocumented
Similarly to `obsinfo` in #155, the docs got a bit left behind when the snapcraft functionality was added.  Adding one or two short paragraphs to `README.md` would be great since it would go a long way towards clarifying what it's for and how it works.  Presumably I'm only asking for ~10 minutes of your time ;-)

aspiers:
This updates the repository's general description, and more thoroughly documents each source service.

Fixes #156.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,85 +1,155 @@
 # tar_scm (OBS source service) [![Build Status](https://travis-ci.org/openSUSE/obs-service-tar_scm.png?branch=master)](https://travis-ci.org/openSUSE/obs-service-tar_scm)
 
-This is an [Open Build Service](http://openbuildservice.org/) source
-service. It uses an SCM client to checkout or update a package by
-creating a source tarball from a source code repository.
+This is the git repository for
+[openSUSE:Tools/obs-service-tar_scm](https://build.opensuse.org/package/show/openSUSE:Tools/obs-service-tar_scm),
+which provides several [source
+services](http://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.source_service.html)
+for the [Open Build Service](http://openbuildservice.org/) which all
+assist with packaging source code from SCM (source code management)
+repositories into tarballs.  The authoritative source is
+https://github.com/openSUSE/obs-service-tar_scm.
 
-It supports many options, e.g. it can adjust resulting tarball
-parameters, include or exclude particular files when creating this
-tarball or generate an rpm changelog from the SCM commit log. For the
-full list of options please see tar_scm.service.in.
+## Services
 
-Apart from various SCM like git, hg, bzr or svn it additionally
+### tar_scm *(deprecated)*
+
+`tar_scm` is the legacy source service used to create a source tarball
+from one of the supported SCM (source code management) tools: `git`,
+`hg`, `svn`, and `bzr`.
+
+`tar_scm` supports many options, e.g. it can adjust resulting tarball
+parameters, include or exclude particular files when creating the
+tarball, or generate an `rpm` changelog from the SCM commit log. For the
+full list of options please see `tar_scm.service.in`.
+
+Apart from various SCM like git, hg, bzr or svn, it additionally
 supports `--url` option that allows you to specify URL of the upstream
 tarball to be downloaded.
 
-This is the git repository for
-[openSUSE:Tools/obs-service-tar_scm](https://build.opensuse.org/package/show/openSUSE:Tools/obs-service-tar_scm).
-The authoritative source is
-https://github.com/openSUSE/obs-service-tar_scm
-
-The service can be used in combination with other services like
+`tar_scm` can be used in combination with other services like
 [download_files](https://github.com/openSUSE/obs-service-download_files),
-[extract_file](https://github.com/openSUSE/obs-service-extract_file),
 [recompress](https://github.com/openSUSE/obs-service-recompress) or
 [set_version](https://github.com/openSUSE/obs-service-set_version)
 e.g. within the [GIT integration](https://en.opensuse.org/openSUSE:Build_Service_Concept_SourceService#Example_2:_GIT_integration)
 workflow.
 
+**`tar_scm` is deprecated in favour of `obs_scm`.**
+
+### obs_scm
+
+`obs_scm` is similar in concept to `tar_scm`, but instead of directly
+generating tarballs, it instead uses the new `obscpio` archive format
+(see below) as an intermediate space-efficient format in which to
+store the sources.
+
+**It is recommended to use `obs_scm` in favour to `tar_scm`**, because
+it provides the following advantages:
+
+1. When you `osc checkout`, you'll also get a local checkout directory
+   within the project directory, inside which you can develop as usual
+   and test your changes with local builds, even without having to
+   commit or push your changes anywhere.
+
+2. It helps to save a *lot* of disk space on the server side,
+   especially when used in continuous integration (e.g. nightly builds
+   and builds of pull requests).
+
+The usual source tarballs can be regenerated from this at build-time
+using the `tar` and `recompress` source services, so no changes to
+`.spec` files are required when switching to `obs_scm` and `obscpio`.
+
+Having said that, it may be more efficient to drop the build-time
+usage of `recompress`, since at build-time `rpmbuild` would decompress
+the same file soon after compressing it.  In this case, only the `tar`
+source service would be used to reconstruct an uncompressed tarball to
+be consumed by the `.spec` file.  However this has the side-effect
+that the resulting `.src.rpm` will contain an uncompressed tarball
+too.  This is not necessarily a problem because `.src.rpm` files are
+compressed anyway, and in fact it may even be *more* efficient to
+avoid double compression (i.e. the sources within the `.src.rpm`, and
+the `.src.rpm` itself both being compressed).  But this depends very
+much on the combination of compression formats used for compression of
+the sources, and for compression of the `.src.rpm`.  Therefore the
+decision whether to use `recompress` will depend on what format is
+desired within the resulting `.src.rpm`, and on the types of
+compression being used for both the tarball and by `rpmbuild` for
+constructing the source rpms.
+
+`obs_scm` additionally generates a file named `<package>.obsinfo`
+which includes useful information from your SCM system, such as the
+name, version number, mtime, and commit SHA1.  This data is then used
+by the `tar` service (see below) to reconstruct a tarball for use by
+`rpmbuild` at build-time, and also by the
+[`set_version`](https://github.com/openSUSE/obs-service-set_version)
+source service in order to set the version in build description files
+such as `.spec` or `.dsc` files.
+
+### tar
+
+The `tar` source service creates a tarball out of a `.obscpio` archive
+and a corresponding `.obsinfo` file which contains metadata about it.
+Typically this service is run at build-time, e.g.
+
+    <service name="tar" mode="buildtime"/>
+
+since storing the `.tar` file in OBS would duplicate the source data
+in the `.obscpio` and defeat the point of using `.obscpio` in the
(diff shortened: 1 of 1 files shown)


aspiers:
@M0ses @adrianschroeter @coolo @dirkmueller I attempted to document today's discussions to the best of my understanding.  I'd be really grateful if you could review this new README carefully to ensure it matches your understanding too.  Thanks!

README.md
```diff
@@ -1,85 +1,135 @@
...
+which provides several [source
+services](http://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.source_service.html)
+for the [Open Build Service](http://openbuildservice.org/) which all
+assist with packaging source code from SCM (source code management)
+repositories into tarballs.  The authoritative source is
+https://github.com/openSUSE/obs-service-tar_scm.
+
+## Services
+
+### tar_scm *(deprecated)*
```
adrianschroeter: hm, not sure if "deprecated" is maybe a bit too harsh .... what about just to move the section to the bottom and mark it as (not recommended) alternative?

README.md
```diff
@@ -1,85 +1,135 @@
...
 tarball to be downloaded.
 
-This is the git repository for
-[openSUSE:Tools/obs-service-tar_scm](https://build.opensuse.org/package/show/openSUSE:Tools/obs-service-tar_scm).
-The authoritative source is
-https://github.com/openSUSE/obs-service-tar_scm
-
-The service can be used in combination with other services like
+`tar_scm` can be used in combination with other services like
 [download_files](https://github.com/openSUSE/obs-service-download_files),
```
adrianschroeter: yes, but they don't have any interaction, so why mention it here?

README.md
```diff
@@ -1,85 +1,135 @@
...
 [extract_file](https://github.com/openSUSE/obs-service-extract_file),
```
adrianschroeter: this one is actually not needed anymore here, because we have the "extract" parameter now. It only makes sense when you used download_url or download_files before. But IMHO out of scope here

README.md
```diff
@@ -1,85 +1,135 @@
...
-is basically the `cpio` `newc` format.
+`obs_scm` is similar in concept to `tar_scm`, but instead of directly
+generating tarballs, it instead uses the new `obscpio` archive format
+(see below) as an intermediate space-efficient format in which to
+store the sources.
+
+**It is recommended to use `obs_scm` in favour to `tar_scm`**, because
+it helps to save a lot of disk space on the server side, especially
+when used in continuous integration (e.g. nightly builds and builds of
+pull requests).
```
adrianschroeter: I would highlight the advantages for the user in first place here: You get a local git checkout directory where you can work as usual inside and test your changes with local builds. Even without commit or pushing them.

It helps us on the server side with disk space as an addition, yes;)

README.md
```diff
@@ -1,85 +1,135 @@
...
+
+The usual source tarballs can be regenerated from this at build-time
+using the `tar` and `recompress` source services, so no changes to
+`.spec` files are required when switching to `obs_scm` and `obscpio`.
+Having said that, it may be more efficient to drop the build-time
+usage of `recompress`, since at build-time `rpmbuild` would decompress
+the same file soon after compressing it.  The decision whether to use
```
adrianschroeter: well, this is only true when you also disable the run of local service or the same compression time would happen just before jumping into the build root

README.md
```diff
@@ -1,85 +1,135 @@
...
+`recompress` will depend on what format is desired within the
+resulting `.src.rpm`, and on the types of compression being used for
+both the tarball and by `rpmbuild` for constructing the source rpms.
+
+`obs_scm` additionally generates a file named `<package>.obsinfo`
+which includes useful information from your SCM system, such as the
+name, version number, mtime, and commit SHA1.  This data is then used
+by the `tar` service (see below) to reconstruct a tarball for use by
```
adrianschroeter: and also by the set_version service to patch the build description files.

README.md
adrianschroeter: I would also mention that the content of the .obscpio archives appear as extracted source for the build process. So you can use any service processing these directories. Or you could also theoretically compile directly inside this trees, if your build tool (rpmbuild/...) supports it.

However, you want usually a tar ball as part of your src.rpm, therefore it gets created via the tar service.

README.md
```diff
@@ -1,85 +1,135 @@
...
 
-### appimage
+### obscpio
 
-The `appimage` source service can be used to fetch sources before
-building an [AppImage app](http://appimage.org/).  It parses an
-`appimage.yml` file for the SCM related parts and executes the
-corresponding services.
+The `obscpio` archive format is the new OBS specific archive format
+based on `cpio` which can be consumed and generated by the [OBS Delta
```
adrianschroeter: well, it is just a cpio in "newc" format, not really an OBS invention. But we expect .obscpio to be in "newc" format for being able to process to store it incremental indeed.

adrianschroeter: ah, just read about the newc below ;) I wouldn't call it "OBS specific archive format" nevertheless ... Just write

The `obscpio` archives are cpio archives in "newc" format which can be stored incremental on the server side thanks to OBS Delta Store

README.md
aspiers: @adrianschroeter Are there still some use cases where `tar_scm` is the recommended choice over `obs_scm`?  If not, wouldn't there only be advantages to deprecating it, and no disadvantages?

README.md
aspiers: Fair point ... OTOH promoting awareness of available source services seems like a good thing.  Is there some central index of source services?  If not, we should create one, and then this README can just say "for other source services, see this index".

README.md
aspiers: True - happy to drop this but I think a central index would be beneficial.

README.md
aspiers: > I would highlight the advantages for the user in first place here: You get a local git checkout directory where you can work as usual inside and test your changes with local builds. Even without commit or pushing them.

Oh yeah, great point thanks :)

> I would also mention that the content of the .obscpio archives appear as extracted source for the build process.

I'm not sure I understand - how is that different to what you said in the previous comment about there being a local git checkout directory?  Are you saying that at build time there are *two* unpacked directories - one with a git checkout (including `.git` metadata) and another one which is the unpacked (obs)cpio?

> So you can use any service processing these directories.

Can you give an example of how that would be used?

> Or you could also theoretically compile directly inside this trees, if your build tool (rpmbuild/...) supports it.

Presumably that would make the .src.rpms useless though?  What would be the advantage?

README.md
aspiers: Oh nice, I didn't realise that - thanks!

README.md
aspiers: Sorry, I'm too stupid to understand what you meant here, so I'm going to need some more help ;-)

The way I see it, the pipeline is:

SCM ---(obs_scm)---> `.obscpio` + `.obsinfo` 
    ---(tar)---> `.tar`
    ---(recompress)---> `.tar.xz`
    ---(rpmbuild)---> unpacked sources

Of these 4 steps, the last 3 have to happen at build-time, right?  And the first step (`obs_scm`) can either happen also at build-time (i.e. server-side), or client-side if it's set to disabled and manually triggered via `osc service disabledrun`.

This part of the README is just suggesting that the `recompress` part can be dropped, which would make `rpmbuild` quicker.  How that is related to the decision whether `obs_scm` is put in disabled mode or not?

README.md
aspiers: Yep good idea thanks - rewording.

aspiers:
@adrianschroeter Thanks a lot for the really helpful and thorough review!  I've made some changes but there are still some open questions due to my lack of understanding.

TheAssassin:
@aspiers other than me not liking the combination of words "AppImage app" (I'd say "build an AppImage" and omit the word "app"), it looks good to me. Though, it seems like the YML documentation needs an update. @adrianschroeter, I'd like to ask you to revise it, you're the one who knows best where we are with it.

...I know it's off-topic, but does anyone know about the current state of AppImage and OpenQA?

aspiers:
@TheAssassin Thanks - dropped "app".

M0ses:
> ...I know it's off-topic, but does anyone know about the current state of AppImage and OpenQA?

If anybody @coolo  should know

README.md
adrianschroeter: osc api /service

shows all services including documentation .....

aspiers: That's an awesome tip! But how is anyone supposed to find out about it?

aspiers: It also assumes that you have direct access to `osc`, whereas documentation can be read even on a phone ...

README.md
aspiers: @adrianschroeter Any chance of a reply to this? ;-) Thanks!

README.md
aspiers: Fixing this in the next push ...

README.md
aspiers: Dropping this in next push ...

README.md
M0ses: @aspiers

obs_scm cannot run at buildtime as it needs online access to fetch the sources from your repository.

Please @adrianschroeter correct me if I`m wrong

AFAIK - recompressing the sources makes building slower - indeed - but source rpms (srpm) should also get smaller - So I would recommend it - even if build is slower and more resources are temporarly used in OBS

aspiers: @M0ses wrote:
> obs_scm cannot run at buildtime as it needs online access to fetch the sources from your repository.

Ah, I think maybe I see the confusion now - I was using the term "build-time" in an imprecise way, to mean multiple different things:

- rpmbuild running in a buildroot in a VM
- source services running server-side prior to rpmbuild (I'm not sure exactly when server-side source services are triggered since I only ever used `mode="disabled"`, but I'm assuming `obs_scm` can run server-side just like `tar_scm` and all other source services - right?)
- when `osc service disabledrun` is manually run from the CLI prior to a client-side `osc build`

I guess really it should only be used to mean the first of these.

> AFAIK - recompressing the sources makes building slower - indeed - but source rpms (srpm) should also get smaller - So I would recommend it - even if build is slower and more resources are temporarly used in OBS

Actually it can be the opposite - double compression can make the srpms bigger!

http://mailman.suse.de/mlarch/SuSE/cloud-devel/2017/cloud-devel.2017.10/msg00090.html

Sorry to any non-SUSE employees reading for the internal-only link. This was also discussed on #opensuse-buildservice FreeNode IRC 2017/10/18 around 3pm (that's probably UTC, not sure) although I couldn't find any public logs of that.

Still confused by what @adrianschroeter meant in his comment though...

README.md
adrianschroeter: yes, we used to have a webui interface showing all options with an editor showing text lines, combo boxes and had drag and drop for ordering. It got dropped unfortunatly during some refactoring.

README.md
adrianschroeter: right, obs_scm runs also server side. 

You are correct regarding disabling the recompression topic, how I think we should mention at least that you get very larger src.rpm then. Maybe suggesting that the use should mark it as NoSource in this situation.

aspiers: > you get very larger src.rpm then

Actually you don't:

```
2017-10-18 15:14:27     adrianS the only valid way is to say you don't care about src.rpms
2017-10-18 15:14:34     coolo   adrianS: they are compressed - but with gzip
2017-10-18 15:14:49     adrianS huh ... new to me ....
```

;-)  But sorry, I wrote some better text yesterday which gives more details around this, but forgot to git push it.  Doing that now ...

README.md
aspiers: Ah, that's a shame - that sounded really nice. I hope it comes back at some point!!

README.md
adrianschroeter: k, just ignore me then ... sorry for confusion

PR #208, opened 2017-10-26
Fix testing

M0ses:
Test at build are currently broken


M0ses:
@aspiers Even if we merged now, your review would be appreciated

M0ses:
THX

PR #209, opened 2017-11-01
Avoid pointless git checkout on clone

bluca:
These 2 commits are extracted from this PR: https://github.com/openSUSE/obs-service-tar_scm/pull/179 given they are independent of the feature being added there, and useful by themselves.

The git clone --no-checkout optimisation can save a lot of I/O and time for large repositories like the kernel - the checkout is pointless as it's immediately followed by another checkout of the requested revision.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -258,7 +258,7 @@ class Git(Scm):
         # between multiple services
         org_clone_dir  = self.clone_dir
         self.clone_dir = self.repodir
-        command = ['git', 'clone']
+        command = ['git', 'clone', '--no-checkout']
         use_reference = True
         try:
             if self.args.package_meta:
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -258,7 +258,7 @@ class Git(Scm):
         # between multiple services
         org_clone_dir  = self.clone_dir
         self.clone_dir = self.repodir
-        command = ['git', 'clone', self.repocachedir, self.url, self.clone_dir]
+        command = ['git', 'clone']
         use_reference = True
         try:
             if self.args.package_meta:
@@ -268,10 +268,10 @@ class Git(Scm):
             pass
 
         if use_reference:
-            command = ['git', 'clone', '--reference', org_clone_dir, self.url,
-                       self.clone_dir]
+            command.extend(['--reference', org_clone_dir, self.url])
         else:
-            command = ['git', 'clone', org_clone_dir, self.clone_dir]
+            command.append(org_clone_dir)
+        command.append(self.clone_dir)
         wdir = os.path.abspath(os.path.join(self.clone_dir, os.pardir))
         self.helpers.safe_run(
             command, cwd=wdir, interactive=sys.stdout.isatty())


TarSCM/scm/base.py
```diff
@@ -210,10 +210,13 @@ def is_sslverify_enabled(self):
 
     def version_iso_cleanup(self, version):
         """Reformat timestamp value."""
-        version = re.sub(r'([0-9]{4})-([0-9]{2})-([0-9]{2}) +'
-                         r'([0-9]{2})([:]([0-9]{2})([:]([0-9]{2}))?)?'
-                         r'( +[-+][0-9]{3,4})', r'\1\2\3T\4\6\8', version)
-        version = re.sub(r'[-:]', '', version)
+        # avoid removing "-" (eg: debian downstream revision) if the version
+        # is not a date
+        if re.search(r'([0-9]{4})-([0-9]{2})-([0-9]{2})', version) is not None:
```
aspiers: You don't actually need the parentheses around `[0-9]{4}` in this line, although it doesn't do any harm :-)

bluca:
> This is great - thanks a lot! Any chance you could add a simple test or two covering the date-or-no-date test? Thanks again!

Sure, added a quick testcase in tests/scm.py

tests/scm.py
```diff
@@ -59,3 +59,13 @@ def test_prep_tree_for_archive(self):
...
+
+    def test_version_iso_cleanup(self):
+        # Avoid get_repocache_hash failure in Scm.__init__
+        self.cli.url = ""
+        scm_base = Scm(self.cli, None)
+        self.assertEqual(scm_base.version_iso_cleanup("2.0.1-3"), "2.0.1-3")
+        self.assertEqual(scm_base.version_iso_cleanup("1"), "1")
+        self.assertEqual(scm_base.version_iso_cleanup("1.54-1.2"), "1.54-1.2")
+        self.assertEqual(scm_base.version_iso_cleanup("2017-01-02 02:23 +10"),
+                         "20170102 0223 +10")
```
bluca: @aspiers these are just some basic tests, including the one that would fail without the fix (first one), let me know if you want to add some more

bluca:
@aspiers ping :-)

bluca:
@aspiers ping/2 :-)

bluca:
As requested by @aspiers on IRC taking out the date formatting commit - the second re.sub is actually needed by RPM builds, where the "-" character is illegal in the version so it has to be removed.
So there's more work to do there as there's a fundamental opposition: for RPM builds it must be removed, for DEB builds it must stay. Fun :-)

aspiers:
Updated title and description accordingly.

aspiers:
@bluca It turns out this broke the case where there is a client-side cache being used.  #228 has the details.  I'm trying if `--no-checkout` actually helps significantly, because checking out a large tree and then switching to another revision is a delta operation and doesn't usually require the whole tree to be changed.  So maybe we can revert this.  Any thoughts?

bluca:
Ah sorry about that! It seemed pretty safe, given the revision is immediately checked out after the clone - I'll have a read at that issue.

In general it does make a big difference for large repositories though - try for example to clone the kernel repository, and then to switch to an old (eg: LTS) branch that is very far behind the head of the default branch - on an I/O challenged machine (like an OBS instance), it can take a few minutes.

bluca:
From the other ticket, it looks like it's due to a weird behaviour of git stash pop, that happens when calling via osc and with a cache directory set - so what about avoiding the --no-checkout if those 2 env vars are set as a solution?

bluca:
Your PR seems even better :-)

PR #211, opened 2017-11-08
Check that parent_tag is in an ancestor of the checked out branch

jengelh:
`git rev-list A..B` returns the set of commits present in B but not
A. Only if A is an ancestor of B will the cardinality of the set be
the same as the forward-running offset from A to B.


jengelh:
```
* 12345678 (tag: 1.1.0) bump version to 1.1.0
| * 23456789 (tag: 1.0.2) bump version to 1.0.2
|/
* 34567890 common base
```
combined with
```
<param name="revision">23456789</param>
<param name="parent-tag">1.1.0</param>
```
is such an insatisfiable combination.

aspiers:
@jengelh Sure, but that doesn't answer my previous question:

> Is it only if parent_tag is specified manually as a parameter?

Emphasis on "only", i.e. could it also happen if `parent_tag` is unspecified?  AFAICS the answer is no, but I wanted to check you agree before we amend the commit message accordingly.  This is important to establish because the code you are modifying can be called even if `parent_tag` is unspecified.  Thanks!

jengelh:
git-describe always uses ancestors (except for --contains, but that's not the point now), if there is one.
If there is no ancestor, the service will die with "No tag present for the ...".
The problem really only can show when the service goes to compute the distance by calling git rev-list.

So no, it does not affect implicit `<parent-tag>` AFAICS.

jengelh:
let me retry.
when parent-tag is implicit, the tag git-describe chooses is already known to be an ancestor. Only when this isnexplicitly overridden wifn a different `parent-tag` in the `_service` file is it possible for it not to be an ancestor.

aspiers:
OK good, so we are agreed :-)

jengelh:
message updated

aspiers:
Thanks for the message update, that looks good now.  I still think my previous comment is relevant though:

>Two other suggestions: I think the code for checking ancestry would benefit from being extracted into a separate function call to avoid `_detect_version_tag_offset` getting too long and unwieldly, and secondly please can we have a unit test for this?

TarSCM/scm/git.py
```diff
@@ -201,11 +201,37 @@ def _detect_version_parent_tag(self, parent_tag, versionformat):  # noqa pylint:
...
         return versionformat
 
+    def check_ancestry(self, parent_tag):
+        rcode, tgcommit = self.helpers.run_cmd(
+            ["git", "rev-parse", parent_tag + "^{}"],
+            self.clone_dir
+        )
+        if rcode:
+            sys.exit("\033[31m`git rev-parse " + parent_tag + "` failed: " +
+                     output + "\033[0m")
```
aspiers: Travis [correctly spotted](https://travis-ci.org/openSUSE/obs-service-tar_scm/jobs/395050912) that `output` is not defined here.

TarSCM/scm/git.py
```diff
@@ -201,11 +201,37 @@ def _detect_version_parent_tag(self, parent_tag, versionformat):  # noqa pylint:
...
+        rcode, head = self.helpers.run_cmd(
+            ["git", "rev-parse", "HEAD"],
+            self.clone_dir
+        )
+        if rcode:
+            sys.exit("\033[31m`git rev-parse HEAD` failed: " +
+                     output + "\033[0m")
```
aspiers: Same problem here.

nmorey:
Just saw this pop in my notification.
My two cents:

The easiest way to check if a commit is an ancestor is git merge-base --is-ancestor
`git merge-base --is-ancestor PARENT_TAG CURRENT_SHA`

Returns 0 if it is an ancestor, 1 if it isn't. 
This was introduced in 1.8 so is probably old enough to be available everywhere

PR #214, opened 2017-12-15
Only use current dir if it actually looks like git (Fixes #202)

### LINKED ISSUE #202: Using CACHEDIRECTORY breaks obs_scm
I have this:

```
$ cat ~/.obs/tar_scm
[tar_scm]
CACHEDIRECTORY="/home/vuntz/.cache/obs/tar_scm"
```

When I do `osc service dr` on some crowbar package, I get:

```
Detected cached repository...
debc21eb6fb2d59daa2584c6c9ebf730717701e3
fatal: destination path '/home/vuntz/ibs/Devel:Cloud:8:Staging/crowbar-init' already exists and is not an empty directory.
ERROR(128): "fatal: destination path '/home/vuntz/ibs/Devel:Cloud:8:Staging/crowbar-init' already exists and is not an empty directory."
Command failed(128): "fatal: destination path '/home/vuntz/ibs/Devel:Cloud:8:Staging/crowbar-init' already exists and is not an empty directory."
Aborting: service call failed:  /usr/lib/obs/service/obs_scm --url git://github.com/crowbar/crowbar-init.git --scm git --filename crowbar-init --versionformat 5.0+git.%ct.%h --revision master --submodules disable --changesgenerate enable --outdir /home/vuntz/ibs/Devel:Cloud:8:Staging/crowbar-init/tmp7XT3Wj.obs_scm.service
```

The failure is actually a git failure, and this is the command that fails (called from `prepare_working_copy`):
```
['git', 'clone', '--reference', '/home/vuntz/.cache/obs/tar_scm/e3a48d27add53c51a83d0846a20a41da527434786c9e78d04b3d45d4ac1a1a27/crowbar-init', 'git://github.com/crowbar/crowbar-init.git', '/home/vuntz/ibs/Devel:Cloud:8:Staging/crowbar-init']
```

I suspect the last argument should probably be a subdirectory of the package dir, not the package dir itself...

dirkmueller:
It isn't entirely clear what this code is supposed to be doing
but it is definitely wrong to assume that its a git repo
if it doesn't contain .git.

### DIFF
--- a/TarSCM/scm/base.py
+++ b/TarSCM/scm/base.py
@@ -147,7 +147,9 @@ class Scm():
         # current work directory to allow the developer to work inside
         # of the git repo and fetch local changes
         if sys.argv[0].endswith("snapcraft") or \
-           (self.args.use_obs_scm and os.getenv('OSC_VERSION')):
+           (self.args.use_obs_scm and
+                os.getenv('OSC_VERSION') and
+                os.path.isdir('.git')):
             self.repodir = os.getcwd()
 
         # construct repodir (the parent directory of the checkout)


TarSCM/scm/base.py
```diff
@@ -147,7 +147,9 @@ def prepare_clone_dir(self):
         # current work directory to allow the developer to work inside
         # of the git repo and fetch local changes
         if sys.argv[0].endswith("snapcraft") or \
-           (self.args.use_obs_scm and os.getenv('OSC_VERSION')):
+           (self.args.use_obs_scm and
+                os.getenv('OSC_VERSION') and
+                os.path.isdir('.git')):
```
aspiers: Do we want to support the case where the developer is in a subdirectory of the top-level working directory?

dirkmueller: I believe it does a change to the toplevel dir beforehand

aspiers: I'll believe you ;-)

PR #217, opened 2018-01-21
RFC: git: use shallow copy to avoid downloading history

pavlix:
I was surprised how slow the download from Git was on my local machine and
realized that the module is attempting to download the whole history just
for creating a tarball from the latest commit.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -60,7 +60,7 @@ class Git(Scm):
     def fetch_upstream_scm(self):
         """SCM specific version of fetch_uptream for git."""
         # clone if no .git dir exists
-        command = ['git', 'clone', self.url, self.clone_dir]
+        command = ['git', 'clone', '--depth', '1', self.url, self.clone_dir]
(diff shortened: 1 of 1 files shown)


aspiers:
Thanks for the reminder that it could possibly be more efficient in the non-caching case - in fact there is an ancient `--history-depth` option which used to do this a long time ago but was neutered even before this git repository was created. But unfortunately the approach proposed here is totally broken, as shown by the failing tests.  If it worked for you I'm guessing it was because you didn't have caching enabled when you tested manually, and you didn't run the test suite referenced in [`CONTRIBUTING.md`](https://github.com/openSUSE/obs-service-tar_scm/blob/master/CONTRIBUTING.md) (please correct me if I'm wrong).

When caching is enabled, your shallow clone would go in the cache, and then a second `git clone --reference` happens which fails because the shallow clone isn't referenceable.  This could maybe be fixed by changing it to use `git clone --reference-if-able`, but that would totally defeat the point of the cache.  So my recommendation is that you just enable the cache, and then the clone is only slow the first time.

PR #224, opened 2018-03-05
[backend] Adding http proxy support

doccaz:
Adding http proxy support.

### DIFF
--- a/TarSCM/changes.py
+++ b/TarSCM/changes.py
@@ -202,6 +202,7 @@ class Changes():
         shutil.move(tmp_fp.name, changes_filename)
 
     def get_changesauthor(self, args):
+        changesauthor = None
         # return changesauthor if given as cli option
         if args.changesauthor:
             return args.changesauthor
@@ -211,7 +212,6 @@ class Changes():
             files = [[os.path.join(os.environ['HOME'], '.oscrc'), False]]
             cfg = Config(files)
 
-            changesauthor = None
             section = cfg.get('general', 'apiurl')
             if section:
                 changesauthor = cfg.get(section, 'email')
--- a/TarSCM/scm/base.py
+++ b/TarSCM/scm/base.py
@@ -277,4 +277,3 @@ class Scm():
 
     def finalize(self):
         self.cleanup()
-        pass
--- a/TarSCM/scm/base.py
+++ b/TarSCM/scm/base.py
@@ -17,6 +17,7 @@ if sys.version_info[0] < 3:
 else:
     import urllib
 
+
 class Scm():
     def __init__(self, args, task):
         # default settings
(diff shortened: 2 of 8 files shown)


TarSCM/changes.py
```diff
@@ -202,6 +202,7 @@ def write_changes(self, changes_filename, changes, version, author):
         shutil.move(tmp_fp.name, changes_filename)
 
     def get_changesauthor(self, args):
+        changesauthor = None
```
M0ses: This looks like a valid bugfix for me, but a bit out of scope of the rest of the commit, correct?

doccaz: Exactly. I did have to fix this in order to successfully test my changes. Perhaps I should have made this a separate fix?

TarSCM/scm/base.py
```diff
@@ -142,6 +148,27 @@ def _calc_repocachedir(self):
...
 
+    def _calc_proxies(self):
+        # check for standard http/https proxy variables
+        #   - http_proxy
+        #   - https_proxy
+        #   - no_proxy
+        httpproxy  = os.getenv('http_proxy')
+        httpsproxy  = os.getenv('https_proxy')
+        noproxy  = os.getenv('no_proxy')
+
```
M0ses: I`m a bit confused about upper/lower case of these variables.
In "/etc/sysconfig/proxy" upper case variables are used.
In "/etc/profile.d/profile.sh" lower case variables are set from the upper case variables
In "/etc/wgetrc" lower case is used 
The curl manpage says:

```
The environment variables can be specified in lower case or upper case. 
The lower case version has precedence. 
http_proxy is an exception as it is only available in lower case.
```

TarSCM/scm/base.py
```diff
@@ -249,4 +276,5 @@ def unlock_cache(self):
             self.lock_file = None
 
     def finalize(self):
+        self.cleanup()
```
M0ses: This also looks like a bugfix which is not related to proxy, correct?

TarSCM/scm/base.py
```diff
@@ -249,4 +276,5 @@ def unlock_cache(self):
...
         pass
```
M0ses: The "pass" statement should be obsolete then.

TarSCM/changes.py
M0ses: Not for me

TarSCM/scm/base.py
doccaz: No. This will call the corresponding cleanup() method for each SCM. Some of them require creating temporary directories/config files which must be removed after it's done.

TarSCM/scm/base.py
doccaz: You're right, this is redundant.

TarSCM/scm/base.py
doccaz: Actually /etc/profile.d/profile.sh reads HTTP[S]_PROXY (uppercase) from /etc/sysconfig/proxy and declares http[s]_proxy (lowercase) on the system. The sysconfig file is YaST-specific. Wget  respects the lower case variant, as curl does.

PR #225, opened 2018-03-07
make installation of scm's optional

M0ses:
Starting with SLE15 mercurial will be no longer supported.
Therefore we decided to make installation of scm's optional and
check if the command is installed and executable at runtime.

Three changes are only to make pylint happy again:

import order of "import yaml" in TarSCM.tasks
"return None" in TarSCM.scm.git line 235
"return None" in TarSCM.scm.svn line 83

sorry for mixing into this commit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
   - "2.7"
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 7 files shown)


M0ses:
@aspiers  I had no idea how to test this new function because it highly depends on the local environment. Maybe you might have a suggestion for me?

PR #229, opened 2018-04-06
fix "with reference" case of Git.prepare_working_copy

frispete:


M0ses:
@frispete

Thanks for your contribution.
From my POV we should investigate, what is the root cause of this problem. I was not able to reproduce the problem, so I would like to have a test case which covers the problem to make sure that we don`t break it again (e.g. when refactoring the code in future).

aspiers:
+1 for including a test case.

PR #233, opened 2018-04-16
fix build for distros not yet supporting Recommends tag

adrianschroeter:


aspiers:
I merged #235 so I guess this is just about handling `Recommends` and can be rebased.

PR #236, opened 2018-05-21
Remove clone_dir if clone fails

M0ses:
Without this patch, when a clone fails (e.g. because of network problems)
git already created the empty directory. The next time, "git clone" gets
called, it complains, that the directory already exists.

This patch catches the raised exception, deletes the clone_dir and raises
the same exception again. This should prevent from any leftovers while broken
"git clone" run.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -85,8 +85,12 @@ class Git(Scm):
         if self.repocachedir:
             command.insert(2, '--mirror')
         wdir = os.path.abspath(os.path.join(self.repodir, os.pardir))
-        self.helpers.safe_run(
-            command, cwd=wdir, interactive=sys.stdout.isatty())
+        try:
+            self.helpers.safe_run(
+                command, cwd=wdir, interactive=sys.stdout.isatty())
+        except SystemExit as exc:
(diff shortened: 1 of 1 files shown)


M0ses:
Some more considerations:

* Already cached directories should be checked with `git branch` (just to make sure that git works inside the cached directory
* We need a valid test case
* I wrote a script for checking a full cache directory on server side (branch vs. fsck?)
* flake8 seems to be broken ATM (`make check`) - needs to be investigated, how the code could pass Travis

M0ses:
Another suggestion:

We should add something like `ENV{BSSERVICE_VERSION}` in bs_service so we can be sure that we run on server side and if `git branch` in a cache dir fails, then we could remove the cache dir and start with a fresh clone.

The reason for this new Variable: 
* I don`t want to remove a directory on a developers machine without asking, because maybe he has  accidentally misconfigured the service. 
* On a server side I would expect caching some kind of "self-healing"

@adrianschroeter, @mlschroe, @aspiers :

 Does this make sense to you?

aspiers:
@M0ses Yes thanks, these all make sense to me.  Please can you submit separate issues for them?

PR #243, opened 2018-07-02
improve handling of corrupt git cache directories

M0ses:
Without this patch, a corrupt cache directory will never get cleaned and
result in broken services until admins runs a cleanup job on the server
side.

With
https://github.com/openSUSE/open-build-service/commit/5d49a2897f597e3db6ca4efc2572bd1a6564e4df
we added a new ENV var "OBS_SERVICE_DAEMON" in bs_service, so we can ensure
that we are running in server context. If server context is detected, broken i
cache will be removed automatically.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -17,11 +17,11 @@ class Git(Scm):
         # is needed here
         scmcmd = ['git']
         if self.httpproxy:
-                scmcmd += ['-c', 'http.proxy=' +
-                           self.httpproxy]
+            scmcmd += ['-c', 'http.proxy=' +
+                       self.httpproxy]
         if self.httpsproxy:
-                scmcmd += ['-c', 'https.proxy=' +
-                           self.httpsproxy]
+            scmcmd += ['-c', 'https.proxy=' +
+                       self.httpsproxy]
         return scmcmd
 
     def switch_revision(self):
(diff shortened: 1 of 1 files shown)


TarSCM/scm/git.py
```diff
@@ -17,11 +17,11 @@ def _get_scm_cmd(self):
         # is needed here
         scmcmd = ['git']
         if self.httpproxy:
-                scmcmd += ['-c', 'http.proxy=' +
```
M0ses: only a fix for indent to make pylint happy again

TarSCM/scm/git.py
```diff
@@ -129,24 +129,37 @@ def fetch_submodules(self):
     def update_cache(self):
         """Update sources via git."""
         # Force origin to the wanted URL in case it switched
```
M0ses: catch exeption
* in case of server context
  * remove cache dir
  * run "fetch_upstream_scm"
* else
  * log info to cleanup manually
  * raise SystemExit

M0ses:
@aspiers Unfortunatly I could not provoke a broken git cache, so

* we have no test case
* I could not test the code to see if works as expected

maybe you have an hint for me, howto provoke a broken git repo, where a fetch would fail?

I would love to add some test cases, but we need this desperately in production because of many user complaints in the last week.

aspiers:
Sorry, I was on FTO.

[**@M0ses**](https://github.com/M0ses) commented on [2 Jul 2018, 08:05 BST](https://github.com/openSUSE/obs-service-tar_scm/pull/243#issuecomment-401691324 "2018-07-02T07:05:10Z - Replied by Github Reply Comments"):
> @aspiers Unfortunatly I could not provoke a broken git cache, so
> 
> * we have no test case
> * I could not test the code to see if works as expected
> 
> maybe you have an hint for me, howto provoke a broken git repo, where a fetch would fail?
> 
> I would love to add some test cases, but we need this desperately in production because of many user complaints in the last week.

The comments from both @adrianschroeter and yourself suggested that you already had known failures in the production OBS instance - so why couldn't we just capture one of those and use it as the test case?

PR #244, opened 2018-07-10
Proxy variables not being filled correctly

doccaz:
Due to the way python deals with subclasses, you need to call the __init__ method from the super class. E.g. it was not calling the __init__ method from the Scm base class. As a consequence, the proxy variables were not being filled up.

### DIFF
--- a/TarSCM/scm/bzr.py
+++ b/TarSCM/scm/bzr.py
@@ -8,6 +8,8 @@ from TarSCM.scm.base import Scm
 class Bzr(Scm):
     scm = 'bzr'
 
+    Scm.__init__
+
     def _get_scm_cmd(self):
         """Compose a BZR-specific command line using http proxies."""
         # Bazaar honors the http[s]_proxy variables, no action needed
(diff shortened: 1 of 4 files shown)


marcus-h:
On 2018-07-10 10:51:02 -0700, doccaz wrote:
> Due to the way python deals with subclasses, you need to call the __init__ method from the super class. E.g. it was not calling the __init__ method from the Scm base class. As a consequence, the proxy variables were not being filled up.
> 
Hmm which exact problem are you trying to solve? From a quick glance,
your patch seems to be a NOP: the "Scm.__init__" is only executed once
when parsing the class definition and just "returns" an unbound method
object (that is, this statement does not call the superclass' __init__
method).

doccaz:
The _calc_proxies() method from base.py is not being called. With this, the appropriate proxy parameters are not included in the command line for each SCM.

It's the same situation described here: https://stackoverflow.com/questions/16634311/why-not-constructor-of-super-class-invoked-when-we-declare-the-object-of-sub-cla

M0ses:
@doccaz 

I tried using the current implementation with git (and of course a http_proxy) and it worked for me.
Could you add some test cases?
For now I close this PR, but further contributions are welcome, if  you could figure out, what did not work for you.

PR #245, opened 2018-07-18
Fixes for OBSCPIO metadata packing

### LINKED ISSUE #185: obs_scm ignores package_meta
The *.obscpio image doesn't contain a `.git` folder.

ItachiSan:
This pull request includes commits regarding the *obs_scm* and *tar* services in order to allow metadata packaging when using the two services together.

The commits have more detailed descriptions of the work done.

Fixes #185.

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -41,11 +41,12 @@ class ObsCpio(BaseArchive):
     def create_archive(self, scm_object, **kwargs):
         """Create an OBS cpio archive of repodir in destination directory.
         """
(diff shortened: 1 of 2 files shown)


ItachiSan:
@adrianschroeter generally speaking yes, although there are a few exceptions, such as CVS-based ArchLinux packages.
They usually use the metadata in order to generate the release number; such approach would be impossible without metadata.
In case of other services that can write a proper release number, they could replace ArchLinux's `pkgver` function.

PR #246, opened 2018-08-10
Remove unnecessary dependency on unittest2

mcepl:
unittest2 is not required since Python 2.6, and we don't support that on
any reasonable platform.


M0ses:
Thanks for your contribution but SLE11 is supported until 2022 and contains python 2.6

PR #248, opened 2018-08-21
check --extract file path for parent dir

M0ses:


lnussel:
use os.path.normpath() for consistency with subdir?

msmeissn:
doing os.path.normpath will already collapse A/../B to B , so remove all .. 

you should probably not called it before checking for ..

M0ses:
normpath should only remove '..' which can be "resolved" in this path and if there is no attempt to access the parent dir, it might be an useless but still valid path definition

Example:

```
>>> import os
>>> # valid (even if useless)
>>> fn = os.path.normpath('foo/../bar')
>>> print(fn)
bar
>>> # invalid
>>> fn = os.path.normpath('foo/../../bar')
>>> print(fn)
../bar
>>> 
```

PR #251, opened 2018-09-19
Bug 1107507

M0ses:


TarSCM/scm/tar.py
```diff
@@ -20,6 +22,8 @@ def fetch_upstream(self):
             self.args.obsinfo,
             "name"
         )
+        if "/" in self.clone_dir:
+            sys.exit("name in obsinfo contains '/'.")
```
mgerstner: Can you also check for slashes in the "version"? It could contain ".." and "/".

TarSCM/scm/base.py
```diff
@@ -270,6 +270,10 @@ def prep_tree_for_archive(self, subdir, outdir, dstname):
              os.path.samefile(os.path.dirname(src), dst)):
             return
 
+        r_path = os.path.realpath(src)
+        if self.clone_dir not in r_path:
```
mgerstner: Maybe

`if not r_path.startswith(self.clone_dir)` 

would be a more complete check?

TarSCM/archive.py
```diff
@@ -33,6 +33,9 @@ def extract_from_archive(self, repodir, files, outdir):
             if not os.path.exists(src):
                 sys.exit("%s: No such file or directory" % src)
 
+            if os.path.islink(src):
+                sys.exit("%s: Extracting symlinks is not permitted" % src)
```
mgerstner: This check looks like it is insufficient when multiple components are specified like `mylink -> /etc`, `extract = "mylink/fstab"`. `os.path.islink` would return `False` although a symlink is involved. An approach using `realpath` or checking each component following `repodir` seems advisable.

TarSCM/scm/git.py
```diff
@@ -321,3 +321,9 @@ def prepare_working_copy(self):
     # no cleanup is necessary for git
     def cleanup(self):
         pass
+
+    def check_url(self):
+        """check if url is a remote url"""
+        if not re.match("^(https?|ftps?|git)://", self.url):
```
gollub: @M0ses , that check seems to be missing ssh. Might not be common, but I guess is often used in private OBS deployments.

M0ses: I did not add it for purpose.
In OBS, ssh should not be used. 
Without special preparation (known_hosts), it will break or services get stuck.
Running within a disabled service via osc might make sense in some cases.
I could add it if there is a special use case for it, but ATM I see no reason for this.

Another note: If I would add it for ssh, also syntax like 
```git@github.com:openSUSE/obs-service-tar_scm```
might be wanted.

As I already tried to point out, I would not recommend ssh as transport protocol for services at all.

gollub: I see your point.

But this particular change is breaking existing private OBS deployments using ssh transport. (Some of my colleagues run into this already when the picked up the latest tar-scm service package build from the openSUSE:Tools repo).

So I would argue to not break the user experience on this, to keep it somewhat backward compatible. Not breaking existing workflows/deployments is more important then making assumption on special preparations which people might miss (e.g. know_hosts).

At least in the past I got in touch with various such private OBS deployments using Git SSH transport. Especially OBS deployments which date back when there was no smart-http Git backend available. Private OBS deployment might have not migrated all the services files for maintenance-releases, from SSH to a HTTPS transport (or might still not offer smart-http Git backends).

I created this PR: https://github.com/openSUSE/obs-service-tar_scm/pull/253 ... which is not covering the ...
   git@github.com:openSUSE/obs-service-tar_scm

... case.


CC' @cschneemann

PR #253, opened 2018-09-24
Support also SSH urls for Git

gollub:
ssh:// URLs are common in private OBS deployments.

Regression got introduced in:
44b3beeaf8663d46467fb3945ddc3d73c65ac452


M0ses:
Thanks for your contribution.

PR #254, opened 2018-09-24
fix regression from 44b3bee

M0ses:
with checking for local repos, usage of ssh compliant url's was no longer
possible.

This fix should reenable those type of urls as described in "man git-clone"

[user@]host.xy:path/to/repo

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -324,6 +324,17 @@ class Git(Scm):
 
     def check_url(self):
         """check if url is a remote url"""
-        if not re.match("^(https?|ftps?|git|ssh)://", self.url):
+
+        # no local path allowed
+        if re.match('^file:', self.url):
             return False
-        return True
+
+        if '://' in self.url:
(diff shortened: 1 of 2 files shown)


gollub:
I can't access bug 1107507, but I guess the context of those changes were to prevent non-authorized users to create tarball from the local filesystems of the machine running the OBS source-service or some sort of hardening. With respect to Git and SSH and maybe some other SCM and transport protocols, there is obviously the risk of peopling trying "root@localhost:/root/devops.git" or something like that. If those changes were all in context of a security fix, maybe adding a security recommendation section in the README to point out that OBS source-service instances should only get credentials (e.g. SSH keys) with restricted access, to prevent unrestricted access of the central OBS filesystems. Or things like that.

I'm not suggesting to try to filter/exclude URLs like:

    git@127.0.0.1:/root/devops.git
    root@localhost:/root/chef-solo-obs.git


... or something like that.

In case that was the original motivation.

PR #260, opened 2018-11-12
Stop using @opensuse.org addresses to indicate a missing address

jengelh:


jengelh:
review reminder? @aspiers

jengelh:
I guess they don't get PR notifications either. The github system is terrible in that regard :-/

jengelh:
> I would even go one step further, we should fail when no email adress is configured.

Waiting for your patch; meanwhile, please approve ;-)

aspiers:
[**@adrianschroeter**](https://github.com/adrianschroeter) [commented](https://github.com/openSUSE/obs-service-tar_scm/pull/260#issuecomment-457179310 "2019-01-24T12:27:02Z - Replied by Github Reply Comments"):
> I would even go one step further, we should fail when no email adress is configured.
> 
> We really want to have only valid adresses where a person (or a group) is behind.

Agreed.

PR #261, opened 2018-11-20
Add support for extract globbing

### LINKED ISSUE #259: extract doesn't support globbing
The `extract` function doesn't support glob/regex, this would be very useful with patches etc.
The `exclude` function uses regex atm.

Example;
```xml
<services>
  <service name="obs_scm">
    <param name="url">https://github.com/openSUSE/obs-service-tar_scm.git</param>
    <param name="scm">git</param>
    <param name="extract">src/*.patch</param>
    <param name="extract">src/*.diff</param>
  </service>
</services>
```

Jellyfrog:
Fixes #259


M0ses:
Thanks for your contribution - see my comments

PR #266, opened 2018-11-28
git backend: fix operation when log.showSignature is enabled in config

mgerstner:
When running the TarSCM service locally and git configuration item
log.showSignature is enabled, then the git backend fails to work
correctly. For example `git log -n1 --pretty=format:%H -- .` will also
output gpg signature validation information in this case. This messes up
the revision parsing.

To fix this pass `-c log.showSignature=False` to all invocations of `git
log` to disable this feature explicitly. The showSignature feature was
introduced in git 2.10 but passing this command line option should also
work on older versions, since unknown configuration items are silently
ignored by git.

Since `detect_version()` does not use the `_log_cmd()` wrapper, this
invocation needs to be adjusted, too, for making the change in the
central place effective in all code paths.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -175,11 +175,10 @@ class Git(Scm):
                 parent_tag,
                 versionformat)
 
-        version = self.helpers.safe_run(
-            self._get_scm_cmd() + ['log', '-n1', '--date=short',
-                                   "--pretty=format:%s" % versionformat],
-            self.clone_dir
-        )[1]
+        version = self._log_cmd(
+            ['-n1', '--date=short', "--pretty=format:%s" % versionformat],
+            None
+        )
         return self.version_iso_cleanup(version)
 
     def _detect_parent_tag(self, args):
@@ -247,7 +246,8 @@ class Git(Scm):
 
     def _log_cmd(self, cmd_args, subdir):
         """ Helper function to call 'git log' with args"""
-        cmd = self._get_scm_cmd() + ['log'] + cmd_args
+        cmd = self._get_scm_cmd() + \
+            ['-c', 'log.showSignature=False', 'log'] + cmd_args
         if subdir:
             cmd += ['--', subdir]
         return self.helpers.safe_run(cmd, cwd=self.clone_dir)[1]
--- a/tests/unittestcases.py
+++ b/tests/unittestcases.py
@@ -61,22 +61,28 @@ class UnitTestCases(unittest.TestCase):
         scm     = Git(self.cli, self.tasks)
         # pylint: disable=unused-variable,protected-access
         new_cmd = scm._log_cmd(['-n1'], '')  # noqa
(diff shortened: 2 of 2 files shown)


mgerstner:
@M0ses I'm not quite sure what you mean. I tried pretty much that: Getting the service to work independently of users' .gitconfig settings and this is done by changing the output of `git log` to the data that is expected.

I you have an idea for a better approach I'll be happy to investigate.

M0ses:
Hmm, I was thinking about something like:
```
TMPHOME=`mktemp -d`;HOME=$TMPHOME git log --pretty=format:"%h|%aD|%d|[%cn] <%ce>";rmdir $TMPHOME
```
But I found no native way to do this in git e.g. via cli option.
I would like to get rid of users settings in config files at all.
I think this would be more valuable than modifing single parameters of user specific settings.

M0ses:
just to make it clear: the log format was just an example - it needs to be adjusted to our needs

mgerstner:
Well when looking at this from a larger perspective like you want to do the following comes to my mind:

- either we're parsing git output here via a well defined interface. Then git should make sure this doesn't break by means of user configuration i.e. this issue here would be a bug in git.
- or we're parsing git output that is not supposed to be machine-readable. Then we're using the wrong git interface in the first place and this would be a larger bug in this source service.

Trying to circumvent the user's configuration completely also sounds hacky. It could have other side effects we're not thinking of yet. Anyways, the git man page touches this topic with this entry:

```
       GIT_CONFIG_NOSYSTEM
	   Whether to skip reading settings from the system-wide $(prefix)/etc/gitconfig file. This environment variable can be used along with $HOME and $XDG_CONFIG_HOME to create a predictable environment for a picky
	   script, or you can set it temporarily to avoid using a buggy /etc/gitconfig file while waiting for someone with sufficient permissions to fix it.
```

M0ses:
>     * either we're parsing git output here via a well defined interface. Then git should make sure this doesn't break by means of user configuration i.e. this issue here would be a bug in git.

Output which is not explicitly in a structured format (like e.g. --json) is (IMHO) never a well defined interface. On the other hand - printing some extra information because of a (not so obvious) configuration option while explicitly specifying the output format (like with --pretty-format) sounds also weird to me.

>     * or we're parsing git output that is not supposed to be machine-readable. Then we're using the wrong git interface in the first place and this would be a larger bug in this source service.
> 
> 
> Trying to circumvent the user's configuration completely also sounds hacky. It could have other side effects we're not thinking of yet. 

I totally agree. Using the python bindings of libgit2 might be a way better approach for this problem, but ...
Unfortunatly we have to ensure that tar_scm/obs_scm works for each distribution we want to build against in OBS and backporting libgit2 to each distro which has not included it might be a lot effort (for a very limited usecase)

> Anyways, the git man page touches this topic with this entry:
> 
> ```
>        GIT_CONFIG_NOSYSTEM
> ```

I think it would be good addition to the $HOME hack as we could avoid systemwide configration

PR #272, opened 2018-12-03
fix bug with unicode filenames in prep_tree_for_archive

M0ses:


M0ses:
merging without review because productive projects are affected

PR #274, opened 2018-12-05
Another attempt to fix unicode filenames in prep_tree_for_archive

M0ses:


M0ses:
Merging without review because package build broken ATM

PR #279, opened 2019-01-25
Error without changesauthor

M0ses:


adrianschroeter:
I am not really a fan of searching and reading the osc config file directly. IMHO it would be cleaner, if we export author and email also via ENV like we do with the rest. IMHO a cleaner interface. Yes, needs a new osc. But should not harm since it is a new functionality ...

M0ses:
Ready to merge now 
New packages should be released after https://github.com/openSUSE/osc/commit/37ca5535ce6a3992f0b649dddd6972b7be07d3b6 has been released to avoid bad user experience.

PR #280, opened 2019-02-12
Drop BuildRequires for bzr

toabctl:
bzr does not have python3 support and on python3 only Distros, this is
a problem.
So drop BuildRequires for bzr and do not run the tests during package
build and also only Suggests bzr for installation.

### DIFF
--- a/dist/obs-service-tar_scm.spec
+++ b/dist/obs-service-tar_scm.spec
@@ -31,7 +31,6 @@ Source:         %{name}-%{version}.tar.gz
 # based distributions
 #Patch0:         0001-Debianization-disable-running-mercurial-tests.patch
 %if %{with obs_scm_testsuite}
-BuildRequires:  bzr
 BuildRequires:  git-core
 BuildRequires:  mercurial
 BuildRequires:  subversion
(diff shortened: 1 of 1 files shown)


adrianschroeter:
please don't change the Recommends and please just skip for distros which don't support mercurial. We still want to keep that tested by default.

toabctl:
> please don't change the Recommends

Done

> and please just skip for distros which don't support mercurial. We still want to keep that tested by default.

? I didn't change anything for hg

toabctl:
@adrianschroeter anything else I need todo here?

adrianschroeter:
actually, I don't see why we should drop bzr testing here when mercurial is still pulling in python2.

What is your goal? Getting rid of install dependencies on Code 15? We should just release a maintenance update of the current state then ...

toabctl:
bsc#1082696 is the goal

adrianschroeter:
On Mittwoch, 13. Februar 2019, 09:15:06 CET Thomas Bechtold wrote:
> bsc#1082696 is the goal

k, I let's continue the discussion there for now.

-- 

Adrian Schroeter
SUSE Linux Products GmbH,  Maxfeldstr. 5, 90409 Nuernberg, Germany
email: adrian@suse.de

M0ses:
@toabctl @adrianschroeter 

v0.10.6 seems to be a good canidate for bsc#1082696 but this changes should be conditional for SUSE >= 15SP1 and openSUSE Leap >= 15.1

toabctl:
I don't plan to work on this. Closing.

PR #293, opened 2019-03-01
enforce bytes for cpio file list

M0ses:


darix:
I dont see **any** explanation on why you are changing it

PR #297, opened 2019-04-04
fix #286: configure a home directory for git

### LINKED ISSUE #286: using git credentials (server side) not possible
It is currently not possible to use server side git credentials. This is required for the scenario where the git server requires a login and OBS should be able to clone repos without the need to include any credentials in the URL of the _service file.

It does not work because the service currently has no $HOME which means git will never find the git global configuration $HOME/.gitconfig and $HOME/.git-credentials.

If I add def _get_scm_cmd(self):
        os.environ["HOME"] = "/usr/lib/obs"
to git.py it works but the problem is that services don't have homes and /usr/lib/obs actually belongs to root (which is a permission problem for locking the credentials file if it resides in this directory) but is the home of all OBS-users, so the global git/credential configuration should be placed somewhere else.

Please allow a server side configuration for git credentials.

dsamx:
Having a home directory allows to store global git settings for the
service if required, for example this allows to store git credentials
if a generally used private git-backend requires a login which the
scm-service should use.
The only other solution would be to store the login credentials in the
_service file which is unacceptable.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -11,6 +11,10 @@ class Git(Scm):
     scm = 'git'
 
     def _get_scm_cmd(self):
+        # define a home directory to be able to configure
+        # server-side git credentials and other git settings.
+        os.environ["HOME"] = "/usr/lib/obs"
+
         """Compose a GIT-specific command line using http proxies"""
         # git should honor the http[s]_proxy variables, but we need to
         # guarantee this, the variables do not work every time


dsamx:
The python 2.7 build creates a warning:
"W: 18, 8: String statement has no effect (pointless-string-statement)"
I don't get it - if I add the os.environ setting then the git credentials work, so it is not a pointless statement.

PR #302, opened 2019-04-25
[dist] spec file: python3 only and multidist

M0ses:
* changes to remove python2 deps for Leap 15.1/SLE15SP1
* improvements for the following distributions
  * Mageia
  * Mandriva (2011)
  * Scientific Linux


coolo:
this breaks all sle15 builds:
https://build.opensuse.org/package/live_build_log/OBS:Server:Unstable:TestGithub:PR-7495/obs-server:obs-bundled-gems/SLE_15/x86_64

M0ses:
@coolo Should be fixed now thanks to  @darix fix.

PR #308, opened 2019-05-03
change order in GNUMakefile to prefer python3

M0ses:
Since our code should be python3 ready now and all major distribution trying to
obsolete unsupported python2 we change the order of detecting python in the
GNUMakefile


darix:
LGTM

PR #314, opened 2019-06-04
[dist] enable python3 in SLE >= 12

M0ses:
We have problems with encoding in python2 on SLE12 based distributions
like:

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 31: ordinal not in range(128)

Switching over to python3 should not be a problem in SLE > 11 as python3 is
already availible and we reduce the maintainance efforts for tar_scm.

### DIFF
--- a/dist/obs-service-tar_scm.spec
+++ b/dist/obs-service-tar_scm.spec
@@ -22,7 +22,7 @@
 %bcond_with    obs_scm_testsuite
 %endif
 
-%if 0%{?suse_version} >= 1500 || 0%{?fedora_version} >= 29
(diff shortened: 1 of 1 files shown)


dist/obs-service-tar_scm.spec
```diff
@@ -22,7 +22,7 @@
 %bcond_with    obs_scm_testsuite
 %endif
 
-%if 0%{?suse_version} >= 1500 || 0%{?fedora_version} >= 29
+%if 0%{?suse_version} >= 1500 || 0%{?fedora_version} >= 29 || 0%{?sle_version} >= 120000
```
adrianschroeter: %if 0%{?suse_version} >= 1315 || 0%{?fedora_version} >= 29

should do the same and include leap 42.x

PR #316, opened 2019-06-04
[dist] python3 also for 42.3

M0ses:


M0ses:
merging as personally discussed with @adrianschroeter

PR #317, opened 2019-06-14
Override LANGUAGE and LC_ALL to force commands run to use english

antlarr:
Having LANGUAGE=es in the environment was making `osc service disabledrun`
fail because the `No local changes to save\n` text couldn't be parsed by
scm/git.py since git was running in spanish as LANGUAGE has precedence over
LANG, so we now set it too to force using english.

Just to be sure we don't have a similar issue in the future with
different encodings being used because LC_ALL has precedence over LANG,
we also override LC_ALL so we're safe.

### DIFF
--- a/TarSCM/helpers.py
+++ b/TarSCM/helpers.py
@@ -26,6 +26,8 @@ class Helpers():
         # like 'git branch'
         env = os.environ.copy()
         env['LANG'] = 'en_US.UTF-8'
+        env['LANGUAGE'] = 'en'
+        env['LC_ALL'] = 'en_US.UTF-8'
 
         proc = subprocess.Popen(cmd,
                                 shell=False,


M0ses:
Just for the record:

https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html

M0ses:
But I would like to have encoding separated from the used language.

TarSCM/helpers.py
```diff
@@ -26,6 +26,8 @@ def run_cmd(self, cmd, cwd, interactive=False, raisesysexit=False):
         # like 'git branch'
         env = os.environ.copy()
         env['LANG'] = 'en_US.UTF-8'
```
dirkmueller: This one is superseded by LC_ALL, you can remove it

olafhering:
obs-service-tar_scm-0.10.10.1566390389.9f923f8 fails to run on python-2.6 from SLE11.

````
[   30s] AttributeError: 'module' object has no attribute 'check_output'
````

scarabeusiv:
This one seem to be solved on master differently.

PR #318, opened 2019-07-06
[dist] Build with python2 for SLE12-SP3

M0ses:
In SLE12 SP3 is no dateutil for python3, only for python3. This changed
with SLE12 SP4.

This additional condition ensures that SP3 builds with python2 and everthing
above builds with python3.

### DIFF
--- a/dist/obs-service-tar_scm.spec
+++ b/dist/obs-service-tar_scm.spec
@@ -22,10 +22,14 @@
 %bcond_with    obs_scm_testsuite
 %endif
 
+%if 0%{?sle_version} && 0%{?sle_version} < 120400
+%bcond_with    python3
+%else
 %if 0%{?suse_version} >= 1315 || 0%{?fedora_version} >= 29
 %bcond_without python3
 %else
 %bcond_with    python3
+%endif
 %endif
 
 # This list probably needs to be extended


dist/obs-service-tar_scm.spec
```diff
@@ -22,11 +22,15 @@
 %bcond_with    obs_scm_testsuite
 %endif
 
+%if 0%{?sle_version} && 0%{?sle_version} < 120400
```
darix: Why not make it

```
#    TW                       || leap
%if 0%{?suse_version} >= 1500 || 0%{?sle_version} >= 120400 || 0%{?fedora_version} >= 29
```

sle_version is set for leap too.  that way we dont have too many %if-s for this

M0ses: In my last PR I enabled python3 for Leap 42.3 (suse_version = 1315) which includes all the required python3 modules but it turned out that SLE12SP3 does not include python3-python-dateutil. Thats why I want to build SLE12SP3 without python3 but keep it enabled in Leap 42.3.

darix: 42.3 is EOL now. we should just drop the support for it IMHO. (might be a bigger discussion to be had)

PR #322, opened 2019-07-24
separate language and encoding

M0ses:
according to the following PR's we have problems with the current locale
settings:

https://github.com/openSUSE/obs-build/pull/507
https://github.com/openSUSE/obs-service-tar_scm/pull/317

This PR trys to separate language and encoding settings.
* Determines language to "C" as we need predictable output of the called scm
  commands like git/hg/svn/bzr
* Introduce new option "encoding" to give users the possibility to overwrite
  the encoding which they used in their repo.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -120,11 +120,11 @@ flake83:
 .PHONY: test
 test:
 	: Running the test suite.  Please be patient - this takes a few minutes ...
-	LANG=en_US.UTF-8 LC_TYPE=en_US.UTF-8 TAR_SCM_TESTMODE=1 PYTHONPATH=. $(PYTHON) tests/test.py 2>&1 | tee ./test.log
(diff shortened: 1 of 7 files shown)


darix:
I would like to point out that we went with UTF-8 encoding everywhere to avoid re-encoding as we pretty much want UTF-8 **everywhere**.

The only problem with Arch Linux is ... there is no package which would just give us the UTF-8 locale as on other distros. the real solution would be

a) set locale.conf run local-gen as shown in the other PR.
b) have a post install script in obs-service-tar_scm which does the same also during install

that would make the en_US.UTF-8 locale available for use without changing the defaults again.

M0ses:
All test cases (incl. docker-services) passed

olafhering:
obs-service-tar_scm-0.10.10.1566390389.9f923f8
````
[   30s] AttributeError: 'module' object has no attribute 'check_output'
````

Issue #326, opened 2019-08-31
locale issue on CentOS 7


matthewdva:
version 0.10.10 has introduced a locale issue when run on CentOS 7.

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe5 in position 1079: invalid continuation byte`
This error occurs in both python2.7 and python3

The attached patch corrects the issue but admittedly may not be the correct approach.

[locale-decode.patch.gz](https://github.com/openSUSE/obs-service-tar_scm/files/3562077/locale-decode.patch.gz)

pseiler:
I can verify this: Here is a build log which reproduces this error: 
https://bauleiterk001.srv.muenchen.de/package/live_build_log/Systemmanagement/mia-scripts/RHEL_7/x86_64

lethliel:
@matthewdva @pseiler 
Thanks for the report. I found the reason. There is `bokmål` in the output of `locale -a` (at least for RHEL7)

Regarding your patch: It works, but i personally don't like "hard coding" the encoding we assume.
So I submitted a PR with another approach. I don't know which approach the maintainer of obs-service-tar_scm likes better. 

@M0ses: Your thoughts?

juliogonzalez:
I can confirm the same problem at CentOS6. #333 fixes it (after applying my patch from #335)

matthewdva:
@lethliel 
Thank you for getting to the root issue.  I too did not like the concept of my own patch.  I did not want report a problem without a providing a possible solution.  For what it is worth I think your PR is much better than mine.

PR #333, opened 2019-10-29
fix decoding for locale -a containg non-ASCII

lethliel:
On RHEL7 locale -a contains bokmål wich cannot be
decoded.

So we just use the bytes-like object and split usind b'\n'

And just encodde tloc for use as an index of aloc

fixes https://github.com/openSUSE/obs-service-tar_scm/issues/326


pseiler:
Thank you very much @lethliel

Fishwaldo:
Can we get a new release out for this? All my CentOS builds are failing due to this.

PR #338, opened 2019-12-13
Add git-lfs support

jjacque:
[git-lfs](https://git-lfs.github.com/) is an "easy" to store large files in git(hub) repository.
This PR add support for it the same way it's done with git submodules.
```
<services>
  <service name="obs_scm">
    <param name="url">https://github.com/openSUSE/obs-service-tar_scm.git</param>
    <param name="scm">git</param>
    <param name="lfs">enable</param>
    <param name="exclude">.git</param>
    <param name="revision">master</param>
  </service>
```


jjacque:
Just made the changes you asked and also make git-lfs a recommendation for debian package as it make no sense have it mandatory if people are only using tar_scm.

PR #346, opened 2020-02-28
simplify osc git update case a lot

adrianschroeter:
* fixes failure when revision is not a branch
* just one run, no need for two attempts


adrianschroeter:
sorry, need to do something about the problem of breaking osc usage with explicit revision.

PR #351, opened 2020-03-10
re-enabled W503 in flake8

M0ses:


M0ses:
passed checks, merging because its a minor change and reverts only exclusion of W503 in flake8

PR #352, opened 2020-03-11
Goodbye for bzr

scarabeusiv:
With py2 removal from TW progressing well we need to remove bzr.

And remove py2 runs as all base distributions where one can use latest version of this sw are python3 based only.


scarabeusiv:
I am willing to bet I messed up something so lets see how travis likes it.

debian/control
```diff
@@ -10,8 +10,8 @@ X-Python-Version: >= 2.6
 Package: obs-service-tar-scm
 Architecture: all
 Provides: obs-service-obs-scm, obs-service-tar
-Depends: ${misc:Depends}, ${python:Depends}, bzr, git, subversion, cpio, python-dateutil, python-yaml, locales-all
+Depends: ${misc:Depends}, ${python:Depends}, git, subversion, cpio, python-dateutil, python-yaml, locales-all
 Recommends: mercurial, git-buildpackage, git-lfs, python-keyring, python-keyrings.alt
 Description: An OBS source service: fetches SCM tarballs
  This is a source service for openSUSE Build Service.
- It supports downloading from svn, git, hg and bzr repositories.
+ It supports downloading from svn, git, hg repositories.
```
ismail: "It supports downloading from svn, git and hg repositories."

OlegGirko:
Hey, why did you remove support for `bzr` without considering its [more modern replacement](https://www.breezy-vcs.org/)?

Conan-Kudo:
@OlegGirko Most likely because `breezy` isn't available in any SUSE distribution yet. It was [just added to `devel:tools:scm` last week](https://build.opensuse.org/package/show/devel:tools:scm/breezy). This PR is probably premature, and the bzr functionality could be adapted to use brz instead.

Breezy has been available in Fedora since Fedora 31, and replaces Bazaar in Fedora 32.

debian/control
scarabeusiv: adjusted

scarabeusiv:
I don't see breezy even on a way to Factory so we simply don't have bzr implementation.

Also I didn't check the full compatibility so someone would need to test it even more than just disable this.

Later on the bzr removal is just one commit so it can be reverted and added back with additional tweaks when someone verifies the breezy is up par and properly integrated.

adrianschroeter:
we can consider to drop the support in factory, but this code should remain on older distributions where we maintain it also from git master branch

scarabeusiv:
@adrianschroeter so how exactly should I do that? Just adapt the spec file and leave out the rest of the drop?

Conan-Kudo:
[Breezy is in openSUSE:Factory](https://build.opensuse.org/package/show/openSUSE:Factory/breezy) now, and has been for five months.

PR #354, opened 2020-03-17
Disable gbp by default

M0ses:


nickbroon:
As the depenency on git-buildpackage was only a install time for the specific GBP subpackage, and not a build dependency, why was this needed?
It should be buildable on any release.
obs-service-tar_scm could be installed, but just gbp subpackage would not as install dependency would be missing.
With this change I can no longer get obs-scm-gbp from opensuse:tools :-(

nickbroon:
@M0ses @adrianschroeter any feedback on this?

adrianschroeter:
IIRC we have dependency checks in openSUSE and SUSE distros and they must be fullfillable. Or we won't be able to submit updates there.

So we would need to maintain a list where we can require these dependencies.

nickbroon:
ah, ok understood, thanks.
Can `%enable_gbp` be added to the prj_conf for just opensuse:tools?

adrianschroeter:
On Donnerstag, 30. April 2020, 11:39:36 CEST wrote Nick Brown:
> ah, ok understood, thanks.
> Can `%enable_gbp` be added to the prj_conf for just opensuse:tools?

yes ... done ...

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

PR #355, opened 2020-04-02
fix situation when revision has been modified in _service

adrianschroeter:
to a tag (not a remote branch), but local checkout still follows
a different tag.

The merge just failed, since origin/$tag does not exist. We do
a local merge in that case now to switch to the new revision.


M0ses:
LGTM

PR #357, opened 2020-04-08
Add option exclude-tag which works as a negation of match-tag.

### LINKED ISSUE #221: exclude-tag to filter out tags
Would it be possible to add `exclude-tag` parameter? Similar to `match-tag` but to negate the match. Since git 2.13.0 there is [`--exclude <pattern>`](https://git-scm.com/docs/git-describe/2.13.0#git-describe---excludeltpatterngt).

Why I would need this is to exclude tags that GitHub forcefully creates along with release. I want to avoid `@PARENT_TAG@` to match such tags.

mcepl:
Fixes #221

@aspiers, @noscript Something like this?

### DIFF
--- a/TarSCM/cli.py
+++ b/TarSCM/cli.py
@@ -94,6 +94,9 @@ class Cli():
                             help='Override base commit for @TAG_OFFSET@')
         parser.add_argument('--match-tag',
                             help='tag must match glob(7)')
+        parser.add_argument('--exclude-tag', action='append',
+                            help='Exclude tags identified by glob '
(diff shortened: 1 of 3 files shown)


mcepl:
Also, @adrianschroeter and @M0ses, could I ask for the review here, please? Thank you.

M0ses:
Could you provide some test cases to avoid future regressions?

mcepl:
Replaced-by: https://github.com/openSUSE/obs-service-tar_scm/pull/535

PR #359, opened 2020-04-18
Add @PARENT_TAG@ support for the revision parameter

bluca:
The first reachable tag returned by git describe will be used. Useful to have a _service that always points to the latest release.

Also update debian/changelog which had fallen behind a lot (please try to at least keep the version in sync, actual content is not too important), and improve error message when a command fails.


bluca:
@adrianschroeter and @M0ses any chance you could please have a look, when you got a moment? Thanks!

bluca:
Thanks!

M0ses:
Thanks for your contribution!

PR #362, opened 2020-05-07
dist: Convert to multibuild; run test suite in a separate flavor

DimStar77:
Allows the obs-service-tar_scm package to be used before running
the test suite. This is especially interesting in the case of
distro-bootstrapping, as e.g. python3 does not get all its
dependencies before tar_scm is available.

For Stagings, we basically already have the --without-obs_scm_testing
in place, but that only works partially (due to BuildRequires python3).

### DIFF
--- a/dist/obs-service-tar_scm.spec
+++ b/dist/obs-service-tar_scm.spec
@@ -16,10 +16,22 @@
 #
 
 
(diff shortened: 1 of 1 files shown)


DimStar77:
@M0ses Converted https://build.opensuse.org/request/show/800832

Issue #374, opened 2020-10-09
CentOS 8: nothing provides python-PyYAML, nothing provides python-dateutil, nothing provides python-lxml, nothing provides python >= 2.6


olegantonyan:
Cannot build for CentOS 8: nothing provides python-PyYAML, nothing provides python-dateutil, nothing provides python-lxml, nothing provides python >= 2.6

Substitute: python python3 in project config doesn't help

olegantonyan:
Made a small progress
```
%if 0%{?centos_version} >= 800
Macros:
%_with_python3 1
:Macros
%endif
```
this leads to a new error: `nothing provides python3-base`
Ignore doesn't work once again.

To fix this I had to edit spec file: https://build.opensuse.org/package/rdiff/home:oleg_antonyan/obs-service-tar_scm?opackage=obs-service-tar_scm&oproject=openSUSE%3ATools&rev=2
Basically this line https://github.com/openSUSE/obs-service-tar_scm/blob/master/dist/obs-service-tar_scm.spec#L151 should be removed for CentOS 8

PR #375, opened 2020-10-09
fix build for centos 8 (#374)

olegantonyan:
This fixes OBS build for CentOS 8
Probably needs better solutions, but it works. Example of working package: https://build.opensuse.org/package/show/home:oleg_antonyan/obs-service-tar_scm

Also, this must be in project config:
```
%if 0%{?centos_version} >= 800
Macros:
%_with_python3 1
:Macros
%endif
```

### DIFF
--- a/dist/obs-service-tar_scm.spec
+++ b/dist/obs-service-tar_scm.spec
@@ -150,7 +150,7 @@ BuildRequires:  %{use_python}-lxml
 %if %{with python3}
 
 %if 0%{?centos_version} >= 800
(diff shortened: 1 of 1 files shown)


adrianschroeter:
I am not really a friend to enforce manual changes outside of the package sources. Can't you set the python3 usage inside of the %if block?

adrianschroeter:
should get solved via https://github.com/openSUSE/obs-service-tar_scm/pull/379 . please add a new pull request if I missed something. thanks.

PR #379, opened 2020-11-11
fix build for CentOS adn RHEL 8

adrianschroeter:


M0ses (changes requested):
Output on a centos VM:

```
[kanku@kanku-vm ~]$ rpm --eval '%{?rhel}'
8
[kanku@kanku-vm ~]$ rpm --eval '%{?rhel_version}'

[kanku@kanku-vm ~]$ rpm --eval '%{?centos_version}'

[kanku@kanku-vm ~]$ rpm --eval '%{?centos}'
8
```

So I think

```
%if 0%{?fedora}%{?rhel}
```

would be the better way

M0ses (approved):
LGTM

PR #383, opened 2020-12-17
Add new param --describe-candidates

dmulder:
This new parameter allows you to specify how many
candidates to consider to describe the git input.
We encountered a strange merge situation where
using this parameter was necessary.


dmulder:
Could somebody please review this? The default git describe candidates count is breaking the samba build.

PR #342, opened 2020-01-31
SCM/auth

jjacque:
Use python keyring (especially keyrings.alt encrypted file backend) to store/retrieve SCM credentials. The keyring cli has to be used (or keyring lib) to store, ie:

`sudo -H -u obsservicerun XDG_DATA_HOME=/etc/obs/services/tar_scm.d keyring -b keyrings.alt.file.EncryptedKeyring set https://github.com/jjacque/obs-service-tar_scm.git user1`
This command will interactively prompt you to set the keyring passphrase if it's called for the first time and prompt you for the password you want to store for the service/user combinaison.

It allows users to specify the user to be used in _service file with:
`<param name="user">user1</param>`
`<param name="keyring-passphrase">XXX</param>`

It rewrites source url for hg, bzr, git and use --username/--password for svn.
It aims to give an answer to https://github.com/openSUSE/obs-service-tar_scm/issues/286 as the issue is pretty common.

PR #386, opened 2021-01-04
fix indentation issue making auth for svn only enabled when proxy

jjacque:
@matwey noticed an issue with SVN auth: https://github.com/openSUSE/obs-service-tar_scm/pull/342#issuecomment-753465567
Due to an indentation error the user/password command line params are only added when proxy config is enabled.


adrianschroeter:
already fixed in #385

Issue #367, opened 2020-08-08
Undocumented strip of "-" from tags


bnavigator:
[Hypothesis](https://github.com/HypothesisWorks/hypothesis) uses tags in the form of `hypothesis-python-%{version}`. The following service definition does not work:

```xml
<services>
  <service name="obs_scm" mode="disabled">
    <param name="url">https://github.com/HypothesisWorks/hypothesis.git</param>
    <param name="scm">git</param>
    <param name="revision">hypothesis-python-5.23.11</param>
    <param name="versionformat">@PARENT_TAG@</param>
    <param name="versionrewrite-pattern">hypothesis-python-(.*)</param>
    <param name="subdir">hypothesis-python/tests</param>    
    <param name="filename">hypothesis-python-tests</param>    
  </service>  
</services>
```

Apparently the hyphens are stripped already in the versionformat, so the versionrewrite-pattern logic does not see it:

https://github.com/openSUSE/obs-service-tar_scm/blob/15a0df276c186a9cd77447414f4d9a19f0078e9a/TarSCM/scm/base.py#L319-L331

Please document the strip or move it to after the rewrite.

https://github.com/openSUSE/obs-service-tar_scm/blob/15a0df276c186a9cd77447414f4d9a19f0078e9a/tar_scm.service.in#L73-L75
"unhelpful characters" is not specific enough :(

olafhering:
This is broken.

The dashes are wiped before `versionrewrite-pattern` has a chance to see and process them. It is up to the author of the `_service` file to handle git tags with dashes in it.

mar-v-in:
The update behavior breaks existing implementations and also mismatches the documentation. 

https://github.com/openSUSE/obs-service-tar_scm/blob/a019e6f/tar_scm.service.in#L83-L85

As "unhelpful characters" are no longer removed, hyphens from the `%cd` git date format end up in the final version string which can result in invalid version strings. The documentation claims that `%cd` results in the date without hyphens which was the previous behavior. Either the documentation should be updated to reflect the current implementation or any other system needs to be established to recover the useful behavior to strip hyphens from date formats.

olafhering:
`%cd.%h` resulted in `20200727.d95eb35`. What is the current result of this format string?

mar-v-in:
Would result in `2020-07-27.d95eb35`.

hramrach:
When using versionrewrite-pattern dashes should no longer be stripped.

PR #388, opened 2021-01-13
Remove code which corrupts versions

olafhering:
A _service file which intends to use @PARENT_TAG@ for the package
version, and which uses versionrewrite-pattern to process the tag
further, will be unable to process the full tag name to extract the
desired parts of it via versionrewrite-replacement. The reason is the
forced removal of all hyphens.

It is up to the author of the _service file how the final package
version is assembled. versionrewrite-pattern and
versionrewrite-replacement give control how this needs to be done.

Remove all of the offending code, which should have been removed along
with the changes that introduced versionrewrite in commit
960d0cebad3bbf86102991eb38c542bbf5a16df4.

Fixes issue #367

Signed-off-by: Olaf Hering <olaf@aepfle.de>

### DIFF
--- a/TarSCM/scm/bzr.py
+++ b/TarSCM/scm/bzr.py
@@ -38,7 +38,7 @@ class Bzr(Scm):
             interactive=sys.stdout.isatty()
         )
 
-    def detect_version(self, args):
+    def detect_version(self, args, truncate_at_hyphen):
         """
         Automatic detection of version number for checked-out BZR repository.
         """
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -231,7 +231,7 @@ class Git(Scm):
                 logging.info("Please fix corrupt cache directory!")
                 raise exc
 
-    def detect_version(self, args):
+    def detect_version(self, args, truncate_at_hyphen):
         """
         Automatic detection of version number for checked-out GIT repository.
         """
@@ -259,7 +259,9 @@ class Git(Scm):
                                    "--pretty=format:%s" % versionformat],
             self.clone_dir
         )[1]
-        return self.version_iso_cleanup(version, debian)
+        if truncate_at_hyphen:
+            version = self.version_iso_cleanup(version, debian)
+        return version
 
     def _detect_parent_tag(self, args=None):
         parent_tag = ''
@@ -308,7 +310,7 @@ class Git(Scm):
 
     def get_timestamp(self):
         data = {"parent_tag": None, "versionformat": "%ct"}
-        timestamp = self.detect_version(data)
+        timestamp = self.detect_version(data, False)
         return int(timestamp)
 
     def get_current_commit(self):
--- a/TarSCM/scm/hg.py
+++ b/TarSCM/scm/hg.py
@@ -79,7 +79,7 @@ class Hg(Scm):
             if re.match('.*no changes found.*', str(exc)) is None:
                 raise
 
-    def detect_version(self, args):
+    def detect_version(self, args, truncate_at_hyphen):
         """
         Automatic detection of version number for checked-out HG repository.
         """
@@ -126,11 +126,13 @@ class Hg(Scm):
         ])
 
         version = self.helpers.safe_run(cmd, self.clone_dir)[1]
-        return self.version_iso_cleanup(version)
+        if truncate_at_hyphen:
+            version = self.version_iso_cleanup(version)
+        return version
 
     def get_timestamp(self):
         data = {"parent_tag": None, "versionformat": "{date}"}
-        timestamp = self.detect_version(data)
+        timestamp = self.detect_version(data, True)
         timestamp = re.sub(r'([0-9]+)\..*', r'\1', timestamp)
         return int(timestamp)
 
--- a/TarSCM/scm/svn.py
+++ b/TarSCM/scm/svn.py
@@ -122,7 +122,7 @@ class Svn(Scm):
             else:
                 raise exc
 
-    def detect_version(self, args):
+    def detect_version(self, args, truncate_at_hyphen):
         """
         Automatic detection of version number for checked-out SVN repository.
         """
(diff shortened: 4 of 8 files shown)


hramrach:
Any progress here?

The failing test is python2 based. Is it even maintained? The failure looks like test setup problem.

olafhering:
I can probably fix this by passing the property not as argument, but stuffing it somewhere else so that the actual consumers can grab it.

Any idea what place that would be?

M0ses:
> The failing test is python2 based. Is it even maintained?

short answer: yes.
obs-service-tar_scm has to run (at least) on all maintained versions *SUSE versions (because of buildtime services).

> The failure looks like test setup problem.

Why? Because of pyhton2 support?

M0ses:
> I can probably fix this by passing the property not as argument, but stuffing it somewhere else so that the actual consumers can grab it.
> 
> Any idea what place that would be?

I`m wondering if this should be something which the user should be able to configure, in this case a new parameter would be needed. @adrianschroeter : What do you think.

BTW: test cases would be useful to understand the problem better.

hramrach:
Also it does not work. I applied the patch and I get version with a hyphen:

Index: mlxbf-bootctl.obsinfo
===================================================================
--- mlxbf-bootctl.obsinfo (revision 338286a3367c7b644f23bce59dabeff9)
+++ mlxbf-bootctl.obsinfo (working copy)
@@ -1,5 +1,5 @@
name: mlxbf-bootctl
-version: 1.16.11
+version: mlxbf-bootctl-1.1-6.11
mtime: 1601841854
commit: b6dd3501668c45020f1e81690cd53aa6e21289b0

Index: mlxbf-bootctl.spec
===================================================================
--- mlxbf-bootctl.spec (revision 338286a3367c7b644f23bce59dabeff9)
+++ mlxbf-bootctl.spec (working copy)
@@ -18,7 +18,7 @@


Name: mlxbf-bootctl
-Version: 1.16.11
+Version: mlxbf-bootctl-1.1-6.11
Release: 0
Summary: User-space driver for Mellanox BlueField SoC
License: BSD-2-Clause

hramrach:
Also it breaks vesionrewrite-pattern

`



<services>
        <service name="obs_scm" mode="disabled">
                <param name="scm">git</param>
                <param name="url">https://github.com/Mellanox/mlxbf-bootctl.git</param>
                <param name="filename">mlxbf-bootctl</param>
                <param name="versionrewrite-pattern">mlxbfbootctl(.*)</param>
        </service>
        <service mode="disabled" name="set_version" />
        <service name="tar" mode="buildtime"/>
</services>
`

olafhering:
What does it break? What is the input, what is the expected output?

The goal is to pass the tag verbatim into `versionrewrite-pattern`.

hramrach:
Nevermind, the tag was previously butchered by removing hyphens and now it is not so the pattern must be amended to also contain the hyphens.

olafhering:
> I`m wondering if this should be something which the user should be able to configure, in this case a new parameter would be needed. @adrianschroeter : What do you think.

You mean a `I really mean it` knob? Usage of `versionrewrite-pattern` and `versionrewrite-replacement` is already a clear instruction what is expected to happen.

M0ses:
> > I`m wondering if this should be something which the user should be able to configure, in this case a new parameter would be needed. @adrianschroeter : What do you think.
> 
> You mean a `I really mean it` knob? Usage of `versionrewrite-pattern` and `versionrewrite-replacement` is already a clear instruction what is expected to happen.

Could you please explain the input and the expected output, in best case by a test case? I think I didn`t get the point of your commit.

olafhering:
It must be possible to extract the individual parts of the verbatim tag. It is not up to tar_scm to decide if a hyphen is bad or not. That is entirely the responsibility of the `_service` file author, and a task of `versionrewrite-pattern` and `versionrewrite-replacement`.

hramrach:
Indeed, it works with this service file:
```
<!-- See https://en.opensuse.org/openSUSE:Build_Service_Concept_SourceService -->
<!-- for more details on the syntax -->

<services>
        <service name="obs_scm" mode="disabled">
                <param name="scm">git</param>
                <param name="url">https://github.com/Mellanox/mlxbf-bootctl.git</param>
                <param name="filename">mlxbf-bootctl</param>
                <param name="versionrewrite-pattern">mlxbf-bootctl-(.*)-(.*)</param>
                <param name="versionrewrite-replacement">\1.\2</param>
        </service>
        <service mode="disabled" name="set_version" />
        <service name="tar" mode="buildtime"/>
</services>
```
Nonetheless, removing hyphens was done to prevent invalid version, and it does not happen automatically anymore. This could be seen as a regression.  Maybe an error should be reported when the resulting version contains a hyphen?

olafhering:
rpm will report an error on bogus input.

hramrach:
> > > I`m wondering if this should be something which the user should be able to configure, in this case a new parameter would be needed. @adrianschroeter : What do you think.
> > 
> > 
> > You mean a `I really mean it` knob? Usage of `versionrewrite-pattern` and `versionrewrite-replacement` is already a clear instruction what is expected to happen.
> 
> Could you please explain the input and the expected output, in best case by a test case? I think I didn`t get the point of your commit.

The problem is that upstream uses Debian style release tags such as 1.1-6 but does not adhere to the Debian policy to include only packaging changes in release (-n) and general changes in version (1.1). Hence the -n is significant. tar_scm further slaps the number of commits since the tag at the end 1.1-6.11 (as requested) and silently removes hyphens without any user control -> 1.16.11. This is clearly bogus. Trivially replacing the hyphen with dot gives valid version 1.1.6.11 but there is no way to achieve that with tar_scm.

hramrach:
> rpm will report an error on bogus input.
Clearly there was an attempt to mitigate this in the service itself before that happens.

M0ses:
I opened another PR how I would solve the problem without the risk of tons of broken packages

https://github.com/openSUSE/obs-service-tar_scm/pull/390

comments are welcome

M0ses:
superseded by https://github.com/openSUSE/obs-service-tar_scm/commit/2fdf897c0579a8586a840ed0f681a610476345df

PR #390, opened 2021-02-02
Versionrewrite pattern disables sanitize version

M0ses:


adrianschroeter (dismissed):

hramrach:
Works for me.

M0ses:
> https://build.opensuse.org/package/live_build_log/home:Andreas_Schwab:glibc/glibc/f/x86_64

I have seen that you adjusted your regex and the issue seems to be done now - sorry for the inconvenience but as already @olafhering pointed out - I would like to avoid an "do what I want" switch in addition. Hopefully this slight regression does not effect to much packages.

PR #391, opened 2021-02-02
Debian packaging for gbp service

nickbroon:
Hopefully an improvement; fixing minor old packaging and lintian warnings, enabling GBP support for debian, dependencies and removing pyc from the the package.


debian/control
```diff
@@ -2,7 +2,7 @@ Source: obs-service-tar-scm
 Section: devel
 Priority: optional
 Maintainer: Daniel Gollub <dgollub@brocade.com>
-Build-Depends: debhelper (>= 7.0.0), python3, python3-dateutil, dh-python
+Build-Depends: debhelper (>= 7.0.0), python3, python-argparse, python3-dateutil, dh-python
```
M0ses: is this still required? argparse is part of the python core modules since python 3.2 and we will not support python2 any longer.

nickbroon: Until Debian Bookworm ships without python2 I believe this still needed. Note this only for the debian package and does not impact the rpm for suse.

nickbroon: With Bookworm now available I've dropped this dependency.

nickbroon:
This should now be mergeable. And the source service at https://build.opensuse.org/package/show/openSUSE:Tools/obs-service-tar_scm rerun to pick up the debian.dsc

PR #392, opened 2021-02-12
Support gitlab/github merge requests

adrianschroeter:
OBS may create branches for each merge request via an webhook.
We need to use the source from this request.

The webhook is always storing the post payload in the _branch_request
file.


M0ses (approved):
LGTM

PR #394, opened 2021-03-10
Add option to not add version to tarball

lnussel:
Allows to eg generate an obscpio with constant name which makes it
easier to update locally (avoids osc rm/osc add)


lnussel:
closes #394

M0ses:
should be obsoleted by https://github.com/openSUSE/obs-service-tar_scm/commit/5ff745087759c32674e85bf2917117f44e0b64c8

PR #399, opened 2021-04-13
param --without-version

M0ses:
With this patch it will be possible to avoid automatic extending
the filename with the version.


lnussel:
:+1:

M0ses:
```
Error while moving from 'cockpit' to 'cockpit-238.1'
```
while building ATM, seems like obs-service-tar needs modification

M0ses:
I could not reproduce the buildtime error while moving the source directory in a clean local setup.

PR #407, opened 2021-05-09
Fix version _none_ generate tarball with '-'

zzndb:


M0ses (approved):
LGTM. Thanks for your contribution.

PR #411, opened 2021-05-27
TarScm: use owner/group root in  .obscpio files

mwilck:
Currently, .obscpio archives contain the actual user/group IDs of the
user creating the archive. This causes generated files to be different
for service runs on different systems. In particular, the .obscpio will
be different between a service run on the server and a local run. This
is unfortunate when an automated build with default service mode is converted
to manual build for submission to a project which doesn't allow server-side
service runs, because the .obscpio will change if the service is run locally.

Fix this by changing owner and group in .obscpio files to root.


M0ses (approved):
LGTM

Issue #419, opened 2021-06-16
dashes not removed from commit date when versionrewrite-pattern used


hramrach:
```
<services>
        <service name="obs_scm" mode="localonly">
                <param name="scm">git</param>
                <param name="url">https://github.com/open-power/secvarctl.git</param>
                <param name="revision">main</param>
                <param name="versionformat">@PARENT_TAG@+git@TAG_OFFSET@_%cd_%h</param>
                <param name="changesgenerate">enable</param>
        </service>
        <service name="tar" mode="localonly" />
        <service name="recompress" mode="localonly">
        <param name="file">*.tar</param>
        <param name="compression">xz</param>
        </service>
        <service name="set_version" mode="localonly" />
</services>
```
Version:        v0.2+git20_20210525_652c950
```
<param name="versionrewrite-pattern">v(.*)</param>
```
Version:        0.2+git20_2021-05-25_652c950

M0ses:
Why should it remove dashes from the commit date?

I will quote here my commit message from https://github.com/openSUSE/obs-service-tar_scm/pull/390/commits/49bb443c6011f0a4da0469e37cbbe7df1f0e4b97

```
If versionrewrite_pattern is used we assume you know what you are doing.
If it is not set we sanitize the version at the end avoid corrupt spec
e.g if version hyphens.
```
We did not change the default behaviour as it might break tons of existing setups - but we disabled sanitizing if versionrewrite_pattern is set because versionrewrite_pattern is an advanced feature for advanced users.

hramrach:
This is not about dashes that are injected from the tag but dashes that are injected from the date by the service itself.

The date is documented as `%cd` `YYYYMMDD` but it may become `YYYYMMDD` or `YYYY-MM-DD` depending on presence of `versionrewrite_pattern`. This behavior does not make any sense whatsoever and is in conflict with documentation.

M0ses:
This is what you might be looking for:

```
                <param name="versionrewrite-pattern">v(.*_)(\d{4})-(\d{2})-(\d{2})(_.*)</param>
                <param name="versionrewrite-replacement">\1\2\3\4\5</param>
```

Keep in mind that our services cannot only be used to prepare RPM packaging. Dashes might be bad in RPM version numbers but not in other formats.

M0ses:
There gets nothing injected. 

[...]

PR #421, opened 2021-06-17
Change date format from short to %Y%m%d.

hramrach:
Since commit 49bb443 ("don`t sanitize version if versionrewrite_pattern
is set") the dashes from the short date format get into the version
string when versionrewrite_pattern is specified.

Change the date format so that no dashes are included in the date to
start with so that we don't need to realy on the sanitization to remove
them again.

Fixes: #419
Fixes: 49bb443 ("don`t sanitize version if versionrewrite_pattern is set")
Signed-off-by: Michal Suchanek <msuchanek@suse.de>

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -252,7 +252,7 @@ class Git(Scm):
             versionformat = self._detect_version_tag_offset(
                 self._parent_tag,
                 versionformat)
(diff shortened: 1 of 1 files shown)


M0ses (approved):
LGTM - SEE my additional notes in https://github.com/openSUSE/obs-service-tar_scm/issues/419

PR #425, opened 2021-07-20
Handle cases where git merge does not work

darix:
This can happen e.g. with

```xml
<param name="revision">somegithash</param>
```


M0ses:
Could you rebase to current master and fix the "line too long" errors?

darix:
Those line too long errors are long fixed but your bot messages dont get resolved on top only at the list at the bottom.

M0ses:
I would like to have a test case here, so we do not break it again

PR #427, opened 2021-08-17
Always use `git reset --hard <revision>`

M0ses:
This patch fixes the problem that the working copy is not resetted to the revision if using tar_scm with `osc`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,15 +47,15 @@ jobs:
         run: pip install -r requirements.txt
 
       - name: Lint with flake8
-        run: flake8
+        run: make flake8
 
       - name: Lint code with pylint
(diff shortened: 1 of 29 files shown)


darix:
just a general question... wouldnt this mean you lose all local edits in a working copy when doing "osc build" with build time services? or do you only call `git reset --hard` on the copy afterwards?

M0ses:
> just a general question... wouldnt this mean you lose all local edits in a working copy when doing "osc build" with build time services? 

No - _stash_and_merge should backup your local changes.

> or do you only call `git reset --hard` on the copy afterwards?

No - tar_scm does not restore your latest code changes - its only taking care that you don't lose something.

darix:
+1

darix (approved):

PR #434, opened 2021-12-03
TarScm (cpio): fix symlink time stamps

mwilck:
Symlinks in .obscpio archives don't have the source time stamp. This
may cause non-reproducible .obscpio files.
Fix this by using the "follow_symlinks" option of os.utime().

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -118,7 +118,7 @@ class ObsCpio(BaseArchive):
         tstamp = self.helpers.get_timestamp(scm_object, args, topdir)
         for name in sorted(cpiolist):
             try:
-                os.utime(name, (tstamp, tstamp))
+                os.utime(name, (tstamp, tstamp), follow_symlinks=False)
             except OSError:
                 pass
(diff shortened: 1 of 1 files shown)


mwilck:
ping…

M0ses:
Pong - Could you provide a test case?

mwilck:
Sorry, this is not how it works. I'm providing a bug fix, and a rather obvious one for that matter. Rather than saying "thank you", you're asking me to invest more work into your project. I have no time for that, so either take it or leave it.

If you need a sample .obscpio file, see https://build.opensuse.org/package/show/Base:System/multipath-tools, which contains

```
lrwxrwxrwx   1 root     root           17 Jan  7 21:55 multipath-tools-0.8.8+13+suse.79c3556f/COPYING -> LICENSES/LGPL-2.0
```

This symlink causes `osc service manualrun` to create a cpio archive with a different checksum every time it's being run, because the time stamp differs. My patch avoids that in a trivial way by using the time stamp of the symlink's target rather than the time stamp of the symlink itself.

PR #437, opened 2022-01-11
fixes for python2.7 compatibility

M0ses:
The construct 'str().encode().decode()' may look ugly but it is
required to work on pyhton 2 _and_ 3


TarSCM/tasks.py
```diff
@@ -126,9 +127,9 @@ def finalize(self):
             # we prefix our own here to be sure to not overwrite user files,
             # if he is using us in "disabled" mode
             new_file = args.outdir + '/_service:snapcraft:snapcraft.yaml'
-            with open(new_file, 'w', encoding='utf-8') as outfile:
-                outfile.write(yaml.dump(self.data_map,
-                                        default_flow_style=False))
+            with io.open(new_file, 'w', encoding='utf-8') as outfile:
+                yml_str = yaml.dump(self.data_map, default_flow_style=False)
```
adrianschroeter, line 131: a comment here about the reason behind encode.decode would be nice

tests/fixtures.py
```diff
@@ -118,8 +119,8 @@ def prep_commit(self, new_rev, subdir=None):
             newly_created.append(subdir)
 
         for fname in ('a', subdir + '/b'):
-            with open(fname, 'w', encoding="UTF-8") as fhn:
-                fhn.write(str(new_rev))
+            with io.open(fname, 'w', encoding="UTF-8") as fhn:
+                fhn.write(str(new_rev).encode("UTF-8").decode("UTF-8"))
```
adrianschroeter, line 123: same here and the other places

adrianschroeter (approved):

Issue #436, opened 2021-12-28
Excess lock files


jengelh:
Cf. https://github.com/openSUSE/osc/issues/262#issuecomment-1001935066
https://github.com/openSUSE/osc/issues/262#issuecomment-1001937336

PR #438, opened 2022-01-13
unlink .lock file after unlocking the cache

M0ses:
Without this patch, the ".lock" file in the working directory does not
get removed.

FIXES: #436

### DIFF
--- a/TarSCM/scm/base.py
+++ b/TarSCM/scm/base.py
@@ -375,6 +375,7 @@ class Scm():
         if self.lock_file and os.path.isfile(self.lock_file.name):
             fcntl.lockf(self.lock_file, fcntl.LOCK_UN)
             self.lock_file.close()
+            os.unlink(self.lock_file.name)
             self.lock_file = None
 
     def finalize(self):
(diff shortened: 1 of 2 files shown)


marcus-h:
On 2022-01-13 03:42:26 -0800, Frank Schreiner wrote:
> diff --git a/TarSCM/scm/base.py b/TarSCM/scm/base.py
> index a927c6a..3f34a6e 100644
> --- a/TarSCM/scm/base.py
> +++ b/TarSCM/scm/base.py
> @@ -375,6 +375,7 @@ class Scm():
>          if self.lock_file and os.path.isfile(self.lock_file.name):
>              fcntl.lockf(self.lock_file, fcntl.LOCK_UN)
>              self.lock_file.close()
> +            os.unlink(self.lock_file.name)

This is racy. For instance, let's consider three tar_scm invocations p1, p2,
p3 (which use the same lock file):

p1 starts
p1 creates and opens the lock file
p1 acquires the lock
p2 starts
p2 opens the (existing) lock file
p2 tries to acquire the lock and waits
p1 releases the lock
p2 acquires the lock
p1 unlinks the file
p3 starts
p3 creates and opens the lock file
p3 acquires the lock (there is no lock on the previously _created_ lock file)

At this point, both p2 and p3 are executing...

PR #447, opened 2022-06-20
do not use python mock

MeggyCal:
it is not needed on python3


M0ses (approved):
LGTM - Thanks for your contribution

PR #462, opened 2023-02-16
README.md: Undeprecate tar_scm

hramrach:
obscpio breaks ability to build packages with rpmbuild.

Until that's fixed obscpio should not be used, and hence tar_scm should not be deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ https://github.com/openSUSE/obs-service-tar_scm.
 
 ## Services
 
-### tar_scm *(deprecated)*
+### tar_scm
 
 `tar_scm` is the legacy source service used to create a source tarball
 from one of the supported SCM (source code management) tools: `git`,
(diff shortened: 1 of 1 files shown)


adrianschroeter:
you can run tar also local eg. via "osc service runall".

It is a bit pointless with the move to git for packaging, but we won't support tar_scm server side generated tar balls on the long run.

hramrach:
Which means that the tarballs need to be generated locally with tar_scm, and it should not be marked as deprecated.

hramrach:
I don't really see the point of creating the tarball on the server side when there is already and obscpio file.

The obscpio file could be used as the source, and rpmextract taught how to extract it.

While it is *not* the source, and the source is not provided the obscpio packages are broken.

hramrach:
> you can run tar also local eg. via "osc service runall".

That's exactly what I am doing, only skipping the redundant obscpio step, and having additional set of sources in the package, yet you are saying it's deprecated.

> It is a bit pointless with the move to git for packaging, but we won't support tar_scm server side generated tar balls on the long run.

?! how is that relevant at all?

PR #465, opened 2023-05-18
Fix Change Tar Archive to use gzip compression

arsenalpoll:
This commit modifies the Tar class in BaseArchive to output gzip-compressed tar files instead of uncompressed tar files. This change improves the efficiency of the archive process by reducing the size of the output files.

This change is backward compatible, as tar and gzip are standard on most systems and gzip decompression is automatically handled by tar when extracting the archive.


adrianschroeter:
the compression is used via an extra service atm, so this would break all these setups selecting the compression.

PR #468, opened 2023-05-21
Feat add gzip compression option for tarball creation

arsenalpoll:
This commit introduces the option to create a gzip compressed tarball when the 'tar.gz' extension is selected. 
The open mode for the tarfile is determined based on the chosen extension.

If the extension is 'tar.gz', a gzip compressed tarball will be created. 
Otherwise, a regular uncompressed tarball will be created.

This allows for greater flexibility when creating tarballs, 
without breaking existing setups that do not use gzip compression.

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -161,7 +161,7 @@ class Tar(BaseArchive):
         args                = kwargs['cli']
         outdir              = args.outdir
         dstname             = kwargs['dstname']
-        extension           = (args.extension or 'tar.gz')
+        extension           = (args.extension or 'tar')
         exclude             = args.exclude
         include             = args.include
         package_metadata    = args.package_meta
@@ -170,6 +170,9 @@ class Tar(BaseArchive):
             args,
             scm_object.clone_dir
         )
+        
+        # Check the extension to set the correct tarfile mode
+        open_mode = "w:gz" if extension == 'tar.gz' else "w"
 
         incl_patterns = []
         excl_patterns = []
@@ -224,7 +227,7 @@ class Tar(BaseArchive):
 
         out_file = os.path.join(outdir, dstname + '.' + extension)
 
-        with tarfile.open(out_file, "w:gz", encoding=enc) as tar:
+        with tarfile.open(out_file, open_mode, encoding=enc) as tar:
             try:
                 tar.add(topdir, recursive=False, filter=reset)
             except TypeError:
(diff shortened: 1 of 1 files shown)


arsenalpoll:
@adrianschroeter  Can you help me to take a look at this pull request? Thanks

adrianschroeter:
I'd like to keep the compression handling in the compress service to avoid a double implementation.

Is there any reason why we should duplicate the code?

arsenalpoll:
> I'd like to keep the compression handling in the compress service to avoid a double implementation.
> 
> Is there any reason why we should duplicate the code?

@adrianschroeter Hi, thank you very much for your reply.

My project uses the `debbuild` tool, and debbuild can use the Spec file to build a Ubuntu software package.

During the use of debbuild, using the default TAR method to compress 'tar.gz', an error will be reported during decompression. 
After testing, it is found that changing to TAR.GZ can solve the problem. Therefore, the tar.gz compression method is added.

This is my use example:  `_service`
```xml
<services>
  <service name="tar_scm">
	<param name="scm">git</param>
	<param name="url">http://gitlab.xx.xx/xxx/xyz.git</param>
	<param name="version">_none_</param>
	<param name="revision">origin/dev</param>
	<param name="filename">abc</param>
	<param name="extension">tar.gz</param>
	<param name="extract">abc.spec</param>
  </service>
</services>
```

hramrach:
> > I'd like to keep the compression handling in the compress service to avoid a double implementation.
> > Is there any reason why we should duplicate the code?
> 
> @adrianschroeter Hi, thank you very much for your reply.
> 
> My project uses the `debbuild` tool, and debbuild can use the Spec file to build a Ubuntu software package.
> 
> During the use of debbuild, using the default TAR method to compress 'tar.gz', an error will be reported during decompression. After testing, it is found that changing to TAR.GZ can solve the problem. Therefore, the tar.gz compression method is added.
> 
> This is my use example: `_service`
> 
```
> <services>
>   <service name="tar_scm">
> 	<param name="scm">git</param>
> 	<param name="url">http://gitlab.xx.xx/xxx/xyz.git</param>
> 	<param name="version">_none_</param>
> 	<param name="revision">origin/dev</param>
> 	<param name="filename">abc</param>
> 	<param name="extension">tar.gz</param>
> 	<param name="extract">abc.spec</param>
>   </service>
  <service name="recompress">
    <param name="file">*.tar</param>
    <param name="compression">gz</param>
  </service>
> </services>
```

mcepl:
What @hramrach said, -1 from me.

PR #484, opened 2024-02-28
Create changes file if none exists

JanZerebecki:
This was noticed in a CI generated package with only a _service file, which now fails because I made rpm error when it finds no changes file in Factory.

Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1220540


JanZerebecki:
Tested manually to work for both cases of existing an non-existing changes file. CI failure unrelated.

bmwiedemann (approved):

adrianschroeter:
erm, you always create a changes file now even not wanted?

That won't solve failures where you don't have a changes file on purpose, but build against factory (without any _service).
IMHO you need to solve this on the other side.

JanZerebecki:
@adrianschroeter This only creates the changes file if it is requested via the option for it.

PR #485, opened 2024-03-05
README.md: Undeprecate tar_scm

hramrach:
obscpio breaks ability to build packages with rpmbuild.

Until that's fixed obscpio should not be used, and hence tar_scm should not be deprecated.

Also drop the note about being able to build the obscpio based packages checked out from OBS. Any package can be built locally with osc build but packages *not* based on obscpio can be additionally built with rpmbuild.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ https://github.com/openSUSE/obs-service-tar_scm.
 
 ## Services
 
-### tar_scm *(deprecated)*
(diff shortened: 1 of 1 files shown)


adrianschroeter:
tar balls just have a way to large impact on our source storage, we won't be able to support packages with large tar balls and updating often. We would need to store entire tar balls forever...

PR #492, opened 2024-05-30
Ignore the stderr output when getting the parent tag

jaimeMF:
On https://github.com/organicmaps/organicmaps a warning is printed and it produces and invalid tag name


adrianschroeter:
hm, not being a fan ... it would make debugging way harder in other cases.

jaimeMF:
> hm, not being a fan ... it would make debugging way harder in other cases. What is the root cause of that error output?

With the repository from https://github.com/organicmaps/organicmaps if I checkout the tag `2024.05.11-3-android` and run the command used by the obs service I get:
```shell
>> git describe --tags --abbrev=0
warning: tag '2024.05.11-3-android' is externally known as '2024.05.11-3-android-rc'
2024.05.11-3-android-rc-0-gd8a071a02319e1d724ae6677c148c57fdb159337
```
So it is just a warning, but since it is included with the tag it produces and error later on.

With the last tag on the OrganicMaps repository it doesn't happen.

PR #498, opened 2024-08-01
[dist] fix shebang substitution in spec for py2

M0ses:


M0ses:
further testing required - DO NOT MERGE

adrianschroeter (approved):

adrianschroeter:
the code change is fine, but is our code really still working with python2?

is there a supported platform which enforces python2 to us?

M0ses:
> the code change is fine, but is our code really still working with python2?
> 
> is there a supported platform which enforces python2 to us?

Taken from https://www.suse.com/lifecycle#suse-linux-enterprise-server-11

```
SUSE Linux Enterprise Server 11 SP4 	15 Jul 2015 	31 Mar 2019 	31 Mar 2022 	31 Mar 2028
```

IIRC SLE 11 does not have python3

PR #499, opened 2024-08-01
[dist] fix BuildRequires for older distros like SLE12

M0ses:


M0ses:
successfully tested in SLE 12 SP4

https://build.opensuse.org/package/live_build_log/home:M0ses:OBS:Server:Unstable/obs-service-tar_scm/SLE_12_SP4/x86_64

merging without further reviews.

PR #507, opened 2025-01-09
debian: depend on brz instead of bzr

bluca:
bzr is a deprecated, transitional empty package, pointingn to brz, since Debian 11


bluca:
Dropping it entirely as per https://github.com/openSUSE/obs-service-tar_scm/pull/505 it's even better of course

M0ses (dismissed):
LGTM, approved

M0ses:
I think I was a bit to fast with approving.

Wouldn't it make sense to use `Suggests:` instead of `Depends` for `brz`?

Even if I'm reluctant to drop the bzr code entirely, I would like get rid of hard dependencies on it (both brz and bzr).

M0ses (approved):

PR #508, opened 2025-01-16
Add --ignore-for-workflow parameter

bluca:
When a _service contains multiple obs_scm entries, the OBS SCM workflow branch_package will cause all of them to be updated to the fork/commit of the workflow. But this is rarely what we want, as for example in the case where the source code to build is in one main repository (that should be overridden), but there are also auxiliary repositories where the packaging recipes are stored, and that should not be updated, as the references will be invalid.

This is making it impossible to use branch_package on PRs for the systemd upstream CI project:

https://build.opensuse.org/project/show/system:systemd

The new parameters allows me to configure the auxiliary sources to be skipped by the branch_package update.

### DIFF
--- a/TarSCM/cli.py
+++ b/TarSCM/cli.py
@@ -234,6 +234,10 @@ class Cli():
                                  '(only used with \'--latest-signed-*\')')
         parser.add_argument('--without-version', default = False,
                             help='Do not add version to output file.')
+        parser.add_argument('--ignore-for-workflow', default = False,
+                            help='Do not override the sources when running OBS'
+                                 'SCM branch_package workflow. Useful for'
+                                 'auxiliary files such as recipes.')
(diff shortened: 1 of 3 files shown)


bluca:
@M0ses gentle ping - is this an acceptable way to solve this problem with the OBS workflow? Otherwise, any other suggestions? Thanks!

bluca:
@M0ses rebased this one too, all green

bluca:
@M0ses gentle ping for a review - TIA

M0ses (approved):
LGTM

M0ses:
Thanks for your contribution and sorry for the delay.

bluca:
> Thanks for your contribution and sorry for the delay.

No worries, thanks!

PR #509, opened 2025-01-18
Add new --extract-rename option

bluca:
Allows extracting a file and renaming it. Useful for Debian builds where the recipe file is in git, but the naming doesn't match OBS expectations: usually git stores debian/control but OBS wants debian.control. Also useful for multibuild, eg, to extract foo.spec to foo_TargetRepository.spec.

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -52,6 +52,24 @@ class BaseArchive():
 
                 shutil.copy2(src, outdir)
 
+    def extract_rename_from_archive(self, repodir, tuples, outdir):
+        """Extract and rename all files directly outside of the archive.
+        """
+        if tuples is None:
+            return
+
+        for pair in tuples:
+            path = os.path.join(repodir, pair.split(':')[0])
+
+            if not os.path.exists(path):
+                sys.exit("%s: No such file or directory" % path)
+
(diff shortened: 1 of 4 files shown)


Werkov:
Isn't this what https://github.com/openSUSE/obs-service-extract_file already does? I see you [actually use it](https://build.opensuse.org/projects/home:bluca:systemd/packages/systemd/files/_service?expand=1)  (It seems to me clearer to handle the renaming together with moving of files.)

bluca:
> Isn't this what https://github.com/openSUSE/obs-service-extract_file already does? I see you [actually use it](https://build.opensuse.org/projects/home:bluca:systemd/packages/systemd/files/_service?expand=1) (It seems to me clearer to handle the renaming together with moving of files.)

That only works on tarballs, so you need to use tar_scm instead of obs_scm, and my understanding is that OBS admins want people to use obs_scm as it's easier on the backend and the storage

M0ses:
@bluca Could you rebase to current master. I would like to see if the tests pass now or what breaks. Unfortunatly the old logs where no longer found.

bluca:
@M0ses done, it was a couple of linter warnings, fixed them, all green now

M0ses (approved):

Issue #520, opened 2025-05-15
tar_scm fails with "sre_constants.error: nothing to repeat at position 34"


mwilck:
Service failure for https://build.opensuse.org/package/show/home:mwilck:suse-module-tools/suse-module-tools:

```
service obs_scm failed:
[...]
2025-05-14 20:53:36,640 COMMAND: ['git', 'rev-parse', 'HEAD']
2025-05-14 20:53:36,647 RESULT(0): 'db5c2837fc8f80349f9d29880158814feebbe52a\n'
2025-05-14 20:53:36,655 COMMAND: ['git', 'log', '-n1', '--date=format:%Y%m%d', '--no-show-signature', '--pretty=format:%ct', '--source', 'master']
2025-05-14 20:53:36,663 RESULT(0): '1747255983'
2025-05-14 20:53:36,663 COMMIT TIMESTAMP: 1747255983 (2025-05-14 20:53:03)
Traceback (most recent call last):
File "/usr/lib/obs/service//obs_scm", line 30, in
main()
File "/usr/lib/obs/service//obs_scm", line 26, in main
TarSCM.run()
File "/usr/lib/obs/service/TarSCM/__init__.py", line 39, in run
task_list.process_list()
File "/usr/lib/obs/service/TarSCM/tasks.py", line 132, in process_list
self.process_single_task(task)
File "/usr/lib/obs/service/TarSCM/tasks.py", line 258, in process_single_task
cli = args
File "/usr/lib/obs/service/TarSCM/archive.py", line 124, in create_archive
for name in self.filter_files(filelist, topdir, args):
File "/usr/lib/obs/service/TarSCM/archive.py", line 77, in filter_files
files = [f for f in files if not re.match(excludes, f)]
File "/usr/lib/obs/service/TarSCM/archive.py", line 77, in
files = [f for f in files if not re.match(excludes, f)]
File "/usr/lib64/python3.6/re.py", line 172, in match
return _compile(pattern, flags).match(string)
File "/usr/lib64/python3.6/re.py", line 301, in _compile
p = sre_compile.compile(pattern, flags)
File "/usr/lib64/python3.6/sre_compile.py", line 562, in compile
p = sre_parse.parse(p, flags)
File "/usr/lib64/python3.6/sre_parse.py", line 855, in parse
p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
[...]

PR #521, opened 2025-05-15
Use fnmatch again for excludes

mwilck:
According to the documentation, the value of the "excludes" parameter for obs_scm is a glob. But since commit 140a127 ("fix excludes for obscpio") the code treats this value as a regular expression, which leads to wrong matching results and may lead to python crashes if the glob expression starts with "*".

Fixes: https://github.com/openSUSE/obs-service-tar_scm/issues/520
Fixes: 140a127 ("fix excludes for obscpio")

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -61,7 +61,7 @@ class BaseArchive():
             incl_arr = [fnmatch.translate(x + '*') for x in args.include]
             includes = re_topdir % (re.escape(topdir), r'|'.join(incl_arr))
         if args.exclude:
-            excl_arr = [x for x in args.exclude]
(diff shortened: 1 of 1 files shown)


mwilck:
This PR breaks `test_tar_exclude_re`.
Not unexpectedly, because my patch replaces re matching by glob matching. Commit cfb2e44 suggests that the change from globbing to re matching was deliberate.

But this contradicts the documentation, and it breaks legitimate code. If matching by regular expression is desired, perhaps that could be achieved with some additional syntax, e.g. `<exclude match="re">...</exclude>` ?

mwilck:
Obsoleted by #525

PR #529, opened 2025-09-11
Remove dependency on six

mcepl:
The main point is removal of the dependency on the `six` module. I have added also one small bug fix (in my opinion).


M0ses (approved):
LGTM

PR #545, opened 2026-06-06
Allow specifying multiple versionrewrite-pattern/versionrewrite-replacement

bluca:
In most packaging versioning schemes, like deb, `'-'` has special meaning. `'~'` is typically used to separate RC versions, so that they sort LOWER than the final version and package upgrades work. But in git `'~'` is not a valid character in a git tag, so it cannot be part of a version, so typically `'-'` is used.

IOW, given a commonly used version scheme:

 v123-rc1
 v123

it is currently not possible to both remove the 'v' and optionally substitute the `'-'` with a `'~'`. But to get a valid packaging version from such tags, both need to happen. The issue is that in a single version replacement is not possible to have an optional substitution ('-' -> '~').

Allow specifying multiple versionrewrite-pattern and versionrewrite-replacement settings, and pair them in order (enforcing that the number is the same).

This allows having a first replacement to remove the 'v', and a second one to change `'-rc'` to `'~rc'` that will simply be skipped when the tag is not an rc.

### DIFF
--- a/TarSCM/cli.py
+++ b/TarSCM/cli.py
@@ -43,6 +43,23 @@ def validate_extract_rename(extract_rename):
     return True
 
 
+def validate_versionrewrite(patterns, replacements):
+    # patterns and replacements are paired by index, so when more than one of
+    # either is given their counts must match
+    patterns_len = len(patterns or [])
+    replacements_len = len(replacements or [])
+
+    if (patterns_len > 1 or replacements_len > 1) and \
+            patterns_len != replacements_len:
+        print('--versionrewrite-pattern and --versionrewrite-replacement '
+              'must be specified the same number of times when more than one '
+              'is given (got %d pattern(s) and %d replacement(s))' %
+              (patterns_len, replacements_len))
+        return False
+
+    return True
+
+
 def check_locale(loc):
     try:
         aloc_tmp = subprocess.check_output(['locale', '-a'])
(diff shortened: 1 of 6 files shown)


bluca:
This is a real issue right now in the systemd `_service` where I have to manually switch patterns between RCs:

```
    <param name="version">_auto_</param>
    <param name="versionformat">@PARENT_TAG@.999+@TAG_OFFSET@+g%h</param>
    <!-- Use the following when a release tag is the most recent one:
    <param name="versionrewrite-pattern">v(.*)</param>
    <param name="versionrewrite-replacement">\1</param> -->
    <!-- Use the following when a release candidate tag is the most recent one: -->
    <param name="versionrewrite-pattern">v(\d+)-rc(\d+)|v(\d+)</param>
    <param name="versionrewrite-replacement">\1~rc\2\3</param>
```

which is not very user friendly, and most importantly it completely breaks the SCM PR workflow, as when I switch the pattern, if the submitter's fork does not have all the tags pushed (github does not sync tags across forks) it breaks running the services and breaks the builds.

bluca:
@dirkmueller gentle ping for a review, as this would solve a bit of pain when using SCM workflow for PRs. Thanks!

M0ses (approved):
LGTM

bluca:
Thanks! Could you please check https://github.com/openSUSE/obs-service-tar_scm/pull/551 too if you have a sec? Also needed for SCM workflow

PR #551, opened 2026-06-27
SCM workflow: fetch tags from upstream repo when building _branch_request

bluca:
If the _service files uses the most recent git tag as the base version,
SCM workflow _branch_request builds will often fail as forks usually do
not carry the release tags, as Github doesn't automatically synchronize
them.

Fetch tags from the upstream repo (as recorded in the _service) when
a _branch_request is handled.

### DIFF
--- a/TarSCM/scm/git.py
+++ b/TarSCM/scm/git.py
@@ -133,6 +133,17 @@ class Git(Scm):
         if lang_bak:
             os.environ['LANG'] = lang_bak
 
+    def fetch_upstream_tags(self):
+        """Fetch release tags from the upstream repository into the clone."""
+        if not self.upstream_url:
+            return
+
+        logging.debug("[fetch_upstream_tags] Fetching tags from '%s'",
+                      self.upstream_url)
+
+        command = self._get_scm_cmd() + ['fetch', '--tags', '--filter=tree:0', self.upstream_url]
+        self.helpers.safe_run(command, cwd=self.clone_dir)
+
     def fetch_upstream_scm(self):
         """SCM specific version of fetch_uptream for git."""
         self.auth_url()
--- a/TarSCM/tasks.py
+++ b/TarSCM/tasks.py
@@ -168,14 +168,23 @@ class Tasks():
             dat = json.load(ofh)
         if dat.get('object_kind') == 'merge_request':
             # gitlab merge request
-            args.url = dat['project']['http_url']
-            rev = dat['object_attributes']['source']['default_branch']
-            args.revision = rev
+            head_url = dat['project']['http_url']
+            head_revision = \
+                dat['object_attributes']['source']['default_branch']
         elif dat.get('action') == 'opened':
             # github pull request
-            args.url = "https://github.com/"
-            args.url += dat['pull_request']['head']['repo']['full_name']
-            args.revision = dat['pull_request']['head']['sha']
+            head_url = "https://github.com/"
+            head_url += dat['pull_request']['head']['repo']['full_name']
+            head_revision = dat['pull_request']['head']['sha']
+        else:
+            return args
+
+        # keep the upstream repository for fetching its release tags as the
+        # request head fork may be missing them and they are needed for
+        # version detection via @PARENT_TAG@/@TAG_OFFSET@.
+        args.upstream_url = args.url
+        args.url = head_url
+        args.revision = head_revision
 
         return args
 
--- a/TarSCM/cli.py
+++ b/TarSCM/cli.py
@@ -97,6 +97,7 @@ class Cli():
         self.maintainers_asc = None
         self.url = None
         self.revision = None
+        self.upstream_url = None
         self.user = None
         self.keyring_passphrase = None
         self.changesgenerate = False
@@ -266,6 +267,7 @@ class Cli():
     def verify_args(self, args):
         # basic argument validation
         # pylint: disable=too-many-branches
+        # pylint: disable=too-many-statements
         if not os.path.isdir(args.outdir):
             sys.exit("%s: No such directory" % args.outdir)
 
--- a/TarSCM/scm/base.py
+++ b/TarSCM/scm/base.py
@@ -63,6 +63,8 @@ class Scm():
 
         # optional arguments
         self.revision       = args.revision
+        # used to fetch tags in SCM workflow builds
+        self.upstream_url   = getattr(args, 'upstream_url', None)
         if args.user and args.keyring_passphrase:
             if KEYRING_IMPORT_ERROR == 1:
                 raise SystemExit('Error while importing keyrings.alt.file but '
@@ -166,6 +168,10 @@ class Scm():
         # switch_to_revision
         self.switch_revision()
 
+        # fetch upstream repository tags when building branch request for SCM
+        # workflow
+        self.fetch_upstream_tags()
+
         # git specific: after switching to desired revision its necessary to
         # update
         # submodules since they depend on the actual version of the selected
@@ -180,6 +186,9 @@ class Scm():
         self.unlock_cache()
 
     def fetch_submodules(self):
+        """NOOP in other scm's than git"""
+
+    def fetch_upstream_tags(self):
         """NOOP in other scm's than git"""
 
     def fetch_lfs(self):


hennevogel:
> The expectations with a CI is that a merge commit is built instead, so that the changes being tested are equivalent to what will actually be merged

Just to clarify, no other ci-provider I know of does this. Where does your your "requirement" come from?

bluca:
I thought GHA via the checkout action ran on the target repo, rather than the fork? I might be wrong though

Anyway, the problems this solves are mentioned in the commit message:

- testing what actually is going to be merged rather than what is on the source branch
- lack of git tag syncing breaking scm workflow via obs_scm when package revision is based on the last tag, eg: `<param name="versionformat">@PARENT_TAG@.999+@TAG_OFFSET@+g%h</param>`

I could make this off by default and enabled by a new option if you prefer?

The second issue is causing a lot of spurious CI failures in systemd PRs where I have enabled the workflow for testing, given github doesn't sync tags across forks, the vast majority of PR authors do not have all the git tags in their forks

bluca:
The second problem is also made a lot worse by the limitations of the regex/subst pattern... this other PR mitigates it a lot https://github.com/openSUSE/obs-service-tar_scm/pull/545

hennevogel:
> I thought GHA via the checkout action ran on the target repo, rather than the fork?

That is certainly is not "simulating" a merge commit.

You could argue that `obs_scm` should not substitute the repository from the `_branch_request` request file...

https://github.com/openSUSE/obs-service-tar_scm/blob/991e3a1f2e87c0b611deefb8d95f38a2a25c42a3/TarSCM/tasks.py#L176-L178

but get the commit through the "special" `pull/$NUMBER` refspec (although I'm not sure about gitea/gitlab support here) or a remote. I guess this would solve the problem you subscribe (missing tags).

bluca:
The problem is that I don't have the PR number, I only have the target repo+revision and the source repo+revision. I've changed the implementation, now it's opt-in via a new _service parameter. Defaults to status quo.

hennevogel:
> The problem is that I don't have the PR number... I've changed the implementation

But you could just as easily have the PR number right? :) BTW [gitlab](https://docs.gitlab.com/user/project/merge_requests/merge_request_troubleshooting/#check-out-merge-requests-locally-through-the-head-ref) and gitea also have support for those refspecs...

Or create a remote from the `_branch_request` data and fetch that. There is no need to generate a merge commit.

bluca:
> Or create a remote from the `_branch_request` data and fetch that. There is no need to generate a merge commit.

Ok, I've changed to just fetch the tags from the upstream remote, it solves my main problem so it's good enough for me

hennevogel:
> Ok, I've changed to just fetch the tags from the upstream remote, it solves my main problem so it's good enough for me

Even more elegant, cool.

bluca:
@hennevogel anything else I can do to move this forward? Thanks

hennevogel:
@bluca I'm not merging here so you have to wait for maintainers...

TarSCM/scm/git.py
```diff
@@ -133,6 +133,19 @@ def _stash_and_merge(self):
...
+        """Fetch release tags from the upstream repository into the clone."""
+        if not self.upstream_url:
+            return
+
+        logging.debug("[fetch_upstream_tags] Fetching tags from '%s'",
+                      self.upstream_url)
+
+        command = self._get_scm_cmd() + ['fetch', '--tags', self.upstream_url]
+        if self.partial_clone:
+            command.append('--filter=tree:0')
```
M0ses: Wouldn't it make sense to apply the filter all the time? AFAICS you only want to fetch the tags no tree or blob information.

bluca: Sure, done

M0ses (approved):
From a code perspective it LGTM, but if we would like it as a feature in general I would like to have @adrianschroeter opinion too.

jengelh:
Just chiming in on the tag situation. There are two ways tags can be missing from forks:

* When you fork a Github repo, there is a checkbox "[ ] Fork only the (default) branch". If you use that, that branch will be the only ref that is transferred, excluding branches and excluding tags.
* If you choose to make a full fork, tags are copied only at that particular moment. Any tags created later in the original repository will only trickle down into _working copies_ (when you do something like `git remote update`), but not forks or other arbitrary repositories, and you need to explicitly upload them (e.g. `git push myremote --tags`) instead.

When you switch the `<param name="url">` line in a `_service` file, that's a whole new namespace and there is technically no expectation of seeing the same refnames, let alone the same pointees.

bluca:
> When you switch the `<param name="url">` line in a `_service` file, that's a whole new namespace and there is technically no expectation of seeing the same refnames, let alone the same pointees.

I'm not switching anything. It's the SCM workflow that builds from another fork, as that's the entire premise and purpose of the feature. As explained above, the lack of tags actually breaks obs-service-tar-scm since its `versionformat` parameter no longer works, given there are no tags.

adrianschroeter:
Please note that this is kind of legacy code, since we are about to obsolete these mechanics.

Also the existing usage is currently creating some load issues on our infrastructure. Not because the amount of builds, but because of a relative problematic setup of the projects (eg. the systemd test builds).

I know, it is not your fault, but I'd like to have a better plan first, how to support your valid usecase in future better. Before we add further code on this dead end ...

adrianschroeter:
btw, did you evaluate the pbuild action so that you could build directly on github.com your package as part of your pull requests?

bluca:
> Please note that this is kind of legacy code, since we are about to obsolete these mechanics.

What would be the replacement? If you refer to scmsync, there's no way that can solve the same requirements. It's fine for SUSE-specific package builds or so, but not for generic packages that need to fetch from multiple repositories at the same time and so on.

Regardless, this still exists and is supported, and there's a real issue being fixed here. What would be the problem with fixing it exactly?

> Also the existing usage is currently creating some load issues on our infrastructure. Not because the amount of builds, but because of a relative problematic setup of the projects (eg. the systemd test builds).
> 
> I know, it is not your fault, but I'd like to have a better plan first, how to support your valid usecase in future better. Before we add further code on this dead end ...

What do you mean specifically? I intentionally kept publishing&image building disabled by default in the CI builds, to avoid increasing the load. It's just building packages. Is it because of arm64 builds?

> btw, did you evaluate the pbuild action so that you could build directly on github.com your package as part of your pull requests?

I already have builds on github actions, I don't need pbuild for that. The purpose of the SCM integration is to be able to enable repository publish + image building and publishing with a click when needed, eg: when we need to securely share PR builds with other contributors for advanced testing and so on. It's extremely useful, and nothing of the sorts can be safely done on github actions.

TarSCM/tasks.py
```diff
@@ -168,14 +168,23 @@ def check_for_branch_request(self):
...
+            head_revision = \
+                dat['object_attributes']['source']['default_branch']
         elif dat.get('action') == 'opened':
             # github pull request
-            args.url = "https://github.com/"
-            args.url += dat['pull_request']['head']['repo']['full_name']
-            args.revision = dat['pull_request']['head']['sha']
+            head_url = "https://github.com/"
+            head_url += dat['pull_request']['head']['repo']['full_name']
+            head_revision = dat['pull_request']['head']['sha']
```
AdamMajer, line 178: Just FYI, it's fine to fetch the PR HEAD via target repository, *not* using head repository at all. The SHA will be available directly for fetching there. There is no need to know magic pull numbers here at all, just the SHA of the repository.

The only issue is that the @TAG_OFFSET@ will make no sense here and could be duplicated on for many PRs. But %h will make sure there's unique version.

```python
head_url += dat['pull_request']['base']['repo']['full_name']
head_revision = dat['pull_request']['head']['sha']
```

PR is a branch that is not merged and so @PARENT_TAG@ in reference of another branch makes no sense in this context and is undefined behaviour. @PARENT_TAG@ and @TAG_OFFSET@ will be basically almost random and not indicative of what they are after merging.

bluca: > PR is a branch that is not merged and so @PARENT_TAG@ in reference of another branch makes no sense in this context and is undefined behaviour.

That's just theoretical, and not how it really works in real world usage. In reality, if an scm workflow uses tar-scm with PARENT_TAG in the target repository + branch, it means the fork is off a branch that uses those tags in the first place, so it will have the release tags in the common ancestor. Otherwise the parent repo wouldn't be using PARENT_TAG in the _service in the first place.

PR #553, opened 2026-07-07
Add option to delete source .obscpio archive in tar service

### LINKED ISSUE #515: tar service should have parameter delete to remove the source archive
When I use (I guess preferred) combination of services `obs_scm`, `tar`, `recompress`, I am left with duplicate `*.obscpio` archive. I would like `tar` service cleaning up the source `*.obscpio` after its successful operation.

mcepl:
Introduce a 'delete' parameter for the tar service that allows removing the source `.obscpio` archive after a tarball has been successfully created. This prevents duplicate archives when using the `obs_scm`, `tar`, and `recompress` services in combination.

- Add `--delete` CLI argument (defaults to 'no' for compatibility).
- Implement deletion logic in `Tar.create_archive`.
- Add unit tests in tests/tartests.py to verify deletion behaviour.

Fixes: https://github.com/openSUSE/obs-service-tar_scm/issues/515

### DIFF
--- a/TarSCM/archive.py
+++ b/TarSCM/archive.py
@@ -263,6 +263,12 @@ class Tar(BaseArchive):
 
         self.archivefile    = tar.name
 
+        if args.delete:
+            obscpio_file = os.path.join(outdir, dstname + '.obscpio')
+            if os.path.exists(obscpio_file):
+                logging.debug("Deleting source archive '%s'", obscpio_file)
+                os.remove(obscpio_file)
+
         os.chdir(cwd)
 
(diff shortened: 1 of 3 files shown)


adrianschroeter:
hm, I do not really mind, but why don't you use tar_scm instead of obs_scm & tar   right away?

mcepl:
> hm, I do not really mind, but why don't you use tar_scm instead of obs_scm & tar right away?

Because `tar_scm` is still marked as deprecated?

